PR #1, opened 2022-10-06
GUA 426: Set up the 'write user services' lambda

alex9smith:
This lambda doesn't do anything useful at the moment (it just logs the contents of the event body), but it *should* be deployable and this PR sets up the file structure and a simple unit test which we can build on.

I mostly followed the [AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/typescript-handler.html) and the [Jest getting started guide](https://jestjs.io/docs/getting-started) to get this working, plus I stole a bit of config from [TxMA's event processing lambda](https://github.com/alphagov/di-txma-audit/tree/main/event-processing/lambda). Each commit has more details on what's going on.

I hope all we need to do for the other lambdas we write is copy the config files over to folders, install the same dev dependencies and get coding 🤞🏻 

I went back and forth on the name for this lambda. We called it the 'write user record' lambda in our plan, but that only makes sense until we add a new service in this repo that stores a different kind of user record (eg. the access logs). So I've called it 'write-user-services' which I think should be (more) future proof. Happy to discuss that further.

### DIFF
--- a/lambda/write-user-services/package.json
+++ b/lambda/write-user-services/package.json
@@ -1,0 +1,11 @@
+{
+  "name": "write-user-services",
+  "version": "1.0.0",
+  "description": "A Lambda function to write events to DynamoDB",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "OGL"
+}
--- a/lambda/write-user-services/package.json
+++ b/lambda/write-user-services/package.json
@@ -6,8 +6,12 @@
   "scripts": {
     "test": "jest"
   },
-  "author": "",
+  "author": "GDS",
   "license": "OGL",
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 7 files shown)


lambda/write-user-services/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "write-user-services",
+  "version": "1.0.0",
+  "description": "A Lambda function to write events to DynamoDB",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
```
huwd, line 9: Could do `GDS` or `GDS - One Login for Government`?

huwd (approved):
Great work - sensible decision on the naming things too.
Thanks for all the links in the commits too.
Nice one.

lambda/write-user-services/package.json
alex9smith: Let's go with `GDS` - I'm still not sure what our directorate / programme / product are called 😂

PR #3, opened 2022-10-10
GUA-426: Set up 'happy path' for writing to DynamoDB

alex9smith:
This PR sets up the 'happy path' for the lambda function to write events to DynamoDB. There's a function and test to do data structure validation, but it always returns `true` at the moment. I'll tackle that, error handling and logging in my next PRs.  

I've used the interfaces Huw wrote for the data structures. They're good, but I'm not completely sure about the file structure I've gone with here. I'm also pretty confident that this will work when deployed thanks to the type system, but there might need to be some tweaks when we get to that step, since I can't test deploying it at the moment.

### DIFF
--- a/lambda/write-user-services/models.ts
+++ b/lambda/write-user-services/models.ts
@@ -1,0 +1,13 @@
+type UrnFdnSub = String;
+type ClientId = String;
+
+export interface UserServices {
+  user_id: UrnFdnSub;
+  services: Service[];
+}
+
+export interface Service {
+  client_id: ClientId;
+  count_successful_logins: Number;
+  last_accessed: Date;
+}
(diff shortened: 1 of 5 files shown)


lambda/write-user-services/models.ts
```diff
@@ -0,0 +1,13 @@
+type UrnFdnSub = String;
```
divyaparameswaran, line 1: Do you think we can move the model objects to a common directory as I have a similar structure when writing the format-record lambda ? Apologies , not 100% sure if that would work or not .

divyaparameswaran (approved):

lambda/write-user-services/models.ts
alex9smith: Good question! 

Other teams have been talking about this as well - see [this Slack conversation](https://gds.slack.com/archives/C03UDLGC44W/p1665138690298679) for example. The short answer is yes, but we'll need to make the shared objects into an npm module that gets installed by each of the lambdas.

I don't think that should be too difficult though, so we should definitely do it. I'll have a go in a follow up PR

PR #4, opened 2022-10-10
GUA-426: Handle errors when writing to DynamoDB

alex9smith:
Handle errors when we try to parse an event body and write it to DynamoDB, by allowing them to propagate up to the top-level handler function. This means that if one of the records causes an error, the function can keep processing the rest of the batch. 

For now 'handling the error' just means logging the error message, but this sets us up to do other things like putting the event that caused the error onto a dead letter queue in the future.

This PR also sets up some simple validation rules for the event body - mostly that all the fields are present.

I'd hoped that we'd get this level of checks 'for free' since the parsing function has 
```typescript
return JSON.parse(body) as UserServices;
```

but all the Typescript interfaces get dropped when the program's compiled to Javascript so at run time it can't check the contents of the parsed body against the interface definition. 

---

The structure of the validate functions is interesting. I went with the option of having them return `void` and throwing a `ValidationError` if something goes wrong. 

The other way I thought about structuring it was to have the validation functions return a `boolean` and change the handler function to

```typescript
export const lambdaHandler = async (event: SQSEvent): Promise<void> => {
  for (let i = 0; i < event.Records.length; i++) {
    try {
      const userServices = parseRecordBody(event.Records[i].body);
      if (validateUserServices(userServices) {
        await writeUserServices(userServices);
      } else {
        throw new ValidationError();
      }
    } catch (err) {
      console.error(`ERROR: ${getErrorMessage(err)}`);
    }
  }
};
```
but I felt it was cleaner to keep the error generation logic out of the handler. 

Interestingly, [TxMA's event processing lambda](https://github.com/alphagov/di-txma-audit/blob/main/event-processing/lambda/event-processor-app.ts) has the latter style of error handling for validation errors. I think this is some different language experience coming up - their program structure feels very Java-y with lots of service classes.  My choice to do separate functions and propagate up `ValidationErrors` is more like Python. 

I'm not sure what's correct in JavaScript, and I couldn't see anything useful in the Airbnb style guide, so I'm open to suggestions on the structure here.


alex9smith:
Paul has made some good points about the file structure for lambdas [on Slack](https://gds.slack.com/archives/C011Y5SAY3U/p1665478277559399?thread_ts=1665413961.903389&cid=C011Y5SAY3U): 

> And try to keep everything to one file.
The point being that having more than one file is completely unnecessary for something that only needs 200 lines of code to work, so why split it up?
The purpose of creating objects, classes, methods etc originally was so that these could be shared across multiple files. In a Lambda function, this is not needed.
Also, if you know that most of your functions are single files, you know where most of your code is.
For typescript, having types in a separate file would make sense I suggest since that isn't going to be transpiled to js.

So that makes me feel better about the structure I've chosen to use here; in fact I could even move the validation logic back into `index.js` and get rid of that file. 

What do you think @huwd @divyaparameswaran ?

huwd:
```
return JSON.parse(body) as UserServices;
```
^ that is a really cool tip, didn't know you could return as!

huwd:
Yeah initial thoughts on the comments,
I've been trying to 🤔 the implications of what Paul's been saying, and I think I've ended up in a similar place.

I think that ends me up thinking that we should be quite suspicious of every `import` statement at the top of this file.
My current mental checklist is:
- As few external dependencies as possible, mostly amazon SDK or the occasional dep for some really gnarly data format processing (let's not rewrite a XML parser!)
- Importing interfaces from elsewhere makes sense, they're common to be-reused and shouldn't be in the compiled JS

So yeah, I think your approach here is ticking those boxes!

The only other thing you are importing is Error classes.
Is that because you reckon we'll re-use them, so lets save ourselves writing lots of boilerplate everywhere?

Might want to 🤔 against Paul's recommendations, but seems fine for now.

lambda/write-user-services/index.ts
```diff
@@ -5,9 +5,8 @@ import {
   PutCommand,
   PutCommandOutput,
 } from "@aws-sdk/lib-dynamodb";
-import { UserServices } from "./models";
-import { getErrorMessage } from "./errors";
-import { validateUserServices } from "./validate";
+import { Service, UserServices } from "./models";
+import { getErrorMessage, ValidationError } from "./errors";
```
huwd, line 9: Gotcha, makes more sense now I've seen the commit message too.

huwd (approved):
Great work!

PR #5, opened 2022-10-11
BAU: Set up linting following the Airbnb style guide

alex9smith:
This PR sets up linting for the `write-user-services` lambda function using ESLint and following the Airbnb style guide. 

You can now run the linter with `npm run lint`. IDEs with ESLint and Prettier plugins should also pick up the config file and display errors along with the code.

This isn't used for CI at the moment, but it sets us up to be able to enforce the style guide on PRs in the future.

---

I was hoping to be able to put the `.eslintrc.json` file further up in the file structure, but unfortunately ESLint wasn't happy with that. I think it was looking for a `node_modules/` folder in the same place as the config file which we don't have, and I didn't want to add another place to keep dependencies up to date.

Instead we'll have to copy the config file across to each lambda and install the same dependencies:

```
npm install --save-dev \
  @typescript-eslint/eslint-plugin \
  eslint \
  eslint-config-airbnb-base \
  eslint-config-airbnb-typescript \
  eslint-config-airbnb-typescript-prettier \
  eslint-plugin-import \
  eslint-plugin-prettier \
  eslint-import-resolver-typescript \
  prettier
```

Bit of a faff, but not too bad seeing as we only need to do it once per lambda.

### DIFF
--- a/lambda/write-user-services/.eslintrc.json
+++ b/lambda/write-user-services/.eslintrc.json
@@ -1,0 +1,8 @@
+{
+    "parser": "@typescript-eslint/parser",
+    "extends": [
+        "airbnb-typescript-prettier",
+        "plugin:@typescript-eslint/recommended",
+        "prettier"
+    ]
+}
--- a/lambda/write-user-services/.eslintrc.json
+++ b/lambda/write-user-services/.eslintrc.json
@@ -12,5 +12,8 @@
                 "paths": "./tsconfig.json"
             }
         }
+    },
+    "rules": {
+        "no-console": "off"
     }
 }
--- a/lambda/write-user-services/.eslintrc.json
+++ b/lambda/write-user-services/.eslintrc.json
@@ -14,6 +14,7 @@
         }
     },
     "rules": {
-        "no-console": "off"
+        "no-console": "off",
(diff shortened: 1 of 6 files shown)


lambda/write-user-services/index.ts
```diff
@@ -43,6 +43,7 @@ export const lambdaHandler = async (event: SQSEvent): Promise<void> => {
     try {
       const userServices = parseRecordBody(event.Records[i].body);
       validateUserServices(userServices);
+      /* eslint no-await-in-loop: "off" */
```
huwd, line 46: Would it make more sense to turn it off globally [a config file](https://eslint.org/docs/latest/user-guide/configuring/rules#using-configuration-files)?

I guess it depends if it really is just this case, or if we think this is common enough we'll start seeing this comment turn up around?

I'll approve as non-blocking, but Ideally I reckon we should prefer rules files over inline comments where possible?

huwd (approved):
ace!

lambda/write-user-services/index.ts
alex9smith: I agree that we should block rules in the config if we never want them to apply, but I don't think that's the case here.

The only place I expect this skip command to appear is in the try / catch block in the top level functions. Everywhere else in the lambda we *should* be avoiding `await` in loops, so I decided to keep the rule enabled to cover those.

PR #7, opened 2022-10-11
GUA-428 initial cloudformation

huwd:
Here's a first pass, I think it'd be good to 👀 now and check in on a few points.

Then I'll try applying it to the dev account.
Some questions I have:
- [Code signing config](https://github.com/alphagov/di-txma-audit/blob/bf55421e85a2888f6c3fceeed1f64995ac052853/event-processing/examples/service-integration-example-template.yml#L6-L10), TxMA have that, I suspect we'll need it for the pipeline, no? Can do in a follow up PR if so, let's make this work on dev
- I've not added any logging, this seems like it might be an issue. I note both [TxMA](https://github.com/alphagov/di-txma-audit/blob/bf55421e85a2888f6c3fceeed1f64995ac052853/event-processing/examples/service-integration-example-template.yml#L172-L174) and [Alex](https://github.com/alex9smith/event-processing-lambda/blob/main/infrastructure/template.yaml#L82-L84) have some permissions. Wondering what the relationship is between logging lambdas and the dead letter queues 🤔 could use a chat about that one.
- I've taken some liberties on naming things, let's cross reference with code / folders that the rest of the team are working on. My general names are:
  - QueryServiceRecord
  - FormatServiceRecord
  - WriteServiceRecord
 
I 🤔 about the names a bit, I reckon this is a User's service record, so named things around that. Happy to revise if there's a better name?


infrastructre/template.yaml
```diff
@@ -44,3 +44,11 @@ Resources:
           Value: !Ref 'Environment'
         - Key: Source
           Value: "https://github.com/alphagov/di-account-management-backend/blob/main/infrastructre/template.yaml"
+
+  ######################################
+  # TxMA Event Queue (Dummy for now)
+  ######################################
+  TxMAInputDummyQueue:
+    Type: AWS::SQS::Queue
+    DeletionPolicy: Retain
```
huwd, line 54: cfn-lint wants these to be explicit. Not sure if Retain is right, but figured I'd go with that for now?

alex9smith: I think `Retain` will be annoying when we're testing this in the dev account since we'll end up with lots of queues and Dynamo tables.  We do want it for all the other environments though. I think this is a perfect use case for a `Condition`:

At the top of the file under `Parameters` add in

```
Conditions:
  IsDevelopmentEnvironment: !Equals 
    - !Ref Environment
    - dev
```

Then you can change this (and all the other instances of `Retain`) to 

```suggestion
    DeletionPolicy: !If [IsDevelopmentEnvironment, Delete, Retain]
```

infrastructre/template.yaml
```diff
@@ -52,3 +53,81 @@ Resources:
...
+      CodeUri: ../lambda/query-service-record/query-service-record.ts
+      Events:
+        TxMAInputDummyQueue:
+          Type: SQS
+          Properties: 
+            Queue: !GetAtt TxMAInputDummyQueue.Arn
+      DeadLetterQueue:
+        Type: SQS
+        TargetArn: !GetAtt QueryServiceRecordDeadLetterQueue.Arn
+      Handler: query-service-record.handler
```
alex9smith, line 73: We'll need to check that the handler and file names match what we've put in the lambdas. Won't be much effort to change either way.

alex9smith (commented):
This is a good start! For your questions: 

1. I agree we can wait on the code signing config until we start on the work to deploy this with the pipeline.

2. Logging and the DLQs have different purposes. The DLQs capture events that can't be processed for whatever reason, but we also need to send application logs to Cloudwatch for monitoring response time, memory use etc. For the permissions, we can either add the actions TxMA have, or (preferably) we also give each lambda the AWS managed policy `AWSLambdaBasicExecutionRole` which includes those permissions.
3. Yeah naming is interesting. I called the write lambda `write-user-services` because I think the `record` part is superfluous and that it's clearer to call the object we're creating `UserServices` since it's the collection of services a user has used. So I'd call the three lambdas `QueryUserServices`, `FormatUserServices` and `WriteUserServices` and the DynamoDB table `user_services`. 

Also, minor nitpick but the folder name has a typo - it's `infrastructre` instead of `infrastructure`.

infrastructre/template.yaml
huwd: Nice, I like this! I'll add that!

infrastructre/template.yaml
```diff
@@ -52,3 +53,81 @@ Resources:
...
+    Type: AWS::Serverless::Function
+    Properties:
+      Architectures: ["x86_64"]
+      CodeUri: ../lambda/query-service-record/query-service-record.ts
+      Events:
+        TxMAInputDummyQueue:
+          Type: SQS
+          Properties: 
+            Queue: !GetAtt TxMAInputDummyQueue.Arn
+      DeadLetterQueue:
```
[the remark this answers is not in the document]
alex9smith: (You don't need to do anything differently here, this was me thinking 'out loud')

infrastructre/template.yaml
huwd: What do you think about `UpdateReplacePolicy`? Is that just superfluous?

infrastructre/template.yaml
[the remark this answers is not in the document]
huwd: Yeah sounds good, This is where I'd really like something like tests (... again, what's our integration test stand in here?) to prove what-end-up-where 🤔.

I'll have a closer look at [#4](https://github.com/alphagov/di-account-management-backend/pull/4) when I do the review.

infrastructre/template.yaml
huwd: Cloudformation seems really unhappy with this 🤔 It's saying it's expecting a string.
I've tried:
```
!Join [ '', !If [IsDevelopmentEnvironment, "Delete", "Retain"] ]
```

But it doesn't seem to like that either, I've seen examples of folks doing this on various github issues, so I'm gonna guess this is cfn-lint freaking out.

I reckon we try and apply this and see what SAM says and adjust the linter from there?

infrastructre/template.yaml
```diff
@@ -1,10 +1,271 @@
...
+        Type: SQS
+        TargetArn: !GetAtt QueryServiceRecordDeadLetterQueue.Arn
+      Handler: query-service-record.handler
+      PackageType: Zip
+      MemorySize: 128
+      Runtime: nodejs16.x
+      Role: !GetAtt QueryServiceRecordRole.Arn
+      Timeout: 5
```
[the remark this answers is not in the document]
huwd: 👍🏻

alex9smith (approved):
Great, let's give this a go. I suggest we merge this as-is, merge all the lambda related PRs and then try to deploy it all. We can open a new PR to cover any changes we need to make then.

PR #10, opened 2022-10-13
GUA-426: Simplify `write-user-services` lambda to a single file

alex9smith:
This PR removes the custom error type (which wasn't doing anything) and the parsing function which was just a call to `JSON.parse`. 

This allows us to reduce the lambda to a single file which is in line with the serverless best practices we're following.

---

Don't merge before #9


lambda/write-user-services/tests/write-user-services.test.ts
```diff
@@ -121,7 +112,7 @@ describe("validateUserServices", () => {
 
   describe("throws an error", () => {
     test("when user_id is missing", () => {
-      const userServices = parseRecordBody(
+      const userServices = JSON.parse(
```
huwd, line 114: Can handle in a follow up, but did you see Paul's comment?

I think it was suggesting moving JSON.parsing to it's own function?

Can refactor later if that's what we want to do, will approve for now.

huwd (approved):

lambda/write-user-services/tests/write-user-services.test.ts
alex9smith: I read his comment as the opposite - our parsing function was just calling `JSON.parse` so there was no need for it (which is why I got rid of it here)

PR #12, opened 2022-10-13
GUA-436: Set up CI for Cloudformation and `write-user-services` lambda

alex9smith:
These will run when a pull request is opened and run linting and the unit tests for each lambda, plus lint the Cloudformation template.

To add the other lambdas we'll just need to copy the whole job and change the `working_directory` value for each.

It's possible that the lambda action ends up taking a long time to run, particularly if we end up with a lot of lambdas in this repo, or if some of those lambdas have a lot of dependencies.

If that happens we can look at splitting this action up or caching dependencies [1], but I don't think we need to do that right away.

[1]:https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

### DIFF
--- a/.github/workflows/lint-and-test-lambdas.yaml
+++ b/.github/workflows/lint-and-test-lambdas.yaml
@@ -1,0 +1,21 @@
+name: Lint and test lambdas
+
+on: pull_request
+
+jobs:
+  lint_and_test_write_user_services:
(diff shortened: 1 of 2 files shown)


alex9smith:
The Cloudformation test is currently failing. That's ok, it's picking up valid issues.

PR #14, opened 2022-10-17
 GUA-429 Query user Service lambda , update handler and add tests

divyaparameswaran:
- Lambda to read the events off an SQS queue 
- Query the Dynamo database to get services for the user 
- Create a new User Service record with both the events and send on another SQS queue.
- Add unit test for all the methods.


lambda/query-user-services/query-user-services.ts
```diff
@@ -1 +1,97 @@
...
+
+const dynamoClient = new DynamoDBClient({});
+const dynamoDocClient = DynamoDBDocumentClient.from(
+  dynamoClient,
+  translateConfig
+);
+
+export const parseRecordBody = (body: string): TxmaEvent => {
+  return JSON.parse(body) as TxmaEvent;
+};
```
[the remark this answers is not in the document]
huwd: This is totally take-it-or-leave it, just 🤔 about where we might get duplication with typescript defs

lambda/query-user-services/tests/query-user-services.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    dynamoMock.on(GetCommand).resolves({ Item: serviceList });
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  test("Queries the dynamo db and send an sqs event", async () => {
+    await handler(TEST_SQS_EVENT);
+    expect(sqsMock.commandCalls(SendMessageCommand).length).toEqual(2);
+    expect(dynamoMock.commandCalls(GetCommand).length).toEqual(2);
```
huwd, line 254: Could also check what we're sending there with `toHaveReceivedCommandWith`?
Or you thinking that's covered off in the rest of your test suite?

lambda/query-user-services/query-user-services.ts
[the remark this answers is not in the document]
alex9smith: One of the bits of feedback I got from Paul J was that if our parsing function is just calling out to `JSON.parse` then we don't need a separate parsing function, and I've come around to agree with him.

`JSON.parse` is a well established, stable part of the Javascript core libraries. We can trust that it can take a JSON encoded string and return an object, or throw an error if that string can't be parsed. So I'd argue that we don't need to explicitly unit test the parsing at all (but it's still implicitly covered by the handler tests). If we don't need to unit test the function then I think that's an argument that we don't need a function.

The second argument is that the `as TxmaEvent` or `: TxmaEvent` return types aren't doing anything at runtime - there's no validation that the `any` type object returned by `JSON.parse` has the attributes we expect from the interface since all the types and interfaces are dropped when it's transpiled to Javascript. That's why we still need the validation steps.

So I think we can get rid of this function and change line 88 in the handler function to 
```typescript
const txmaEvent: TxmaEvent = JSON.parse(record.body);
```
and we'll simplify our code and not lose any functionality. 

(also this is the smallest of nitpicks, so if you think I'm being too pedantic then feel free to ignore me)

lambda/query-user-services/query-user-services.ts
```diff
@@ -1 +1,97 @@
...
+
+export const queryUserServices = async (userId: string): Promise<Service[]> => {
+  const command = new GetCommand({
+    TableName: TABLE_NAME,
```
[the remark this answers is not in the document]
alex9smith: Yeah, the `TableName` parameter comes from the [`GetItemCommandInput`](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/interfaces/getitemcommandinput.html) and can be a string or undefined which is why Typescript is ok with it here. 

If it's undefined then the command will fail when it's executed, raise an error which we can catch and handle in the same way as the other errors by putting the event on the DLQ.

lambda/query-user-services/tests/query-user-services.test.ts
divyaparameswaran: Good shout I will include that

lambda/query-user-services/query-user-services.ts
```diff
@@ -1 +1,91 @@
...
+import {
+  SendMessageCommand,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import { DynamoDBDocumentClient, GetCommand } from "@aws-sdk/lib-dynamodb";
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
+
+const { TABLE_NAME, AWS_REGION } = process.env;
+const QUEUE_URL = "";
```
huwd: Ooh I think we need to take the QUEUE_URL in from the environment too?
As this is, we'll pass an empty string to the [`sendSqsMessage` function ](https://github.com/alphagov/di-account-management-backend/pull/14/files#diff-387cce3998b99c626a6e32a1694ae07cd730e4e47ef928d58477c173e31ee1d3R85-R88)and that should make the amazon call a bit sad.

I think if this just goes:
```suggestion
const { TABLE_NAME, AWS_REGION, QUEUE_URL } = process.env;
```

Then you're all good as we'll give it the QUEUE_URL via cloudformation / amazon config.

huwd: Scratch that, i see you just pushed a new commit will review that! 👀

huwd (approved):
Looks great!

PR #15, opened 2022-10-17
GUA-427: Format user services

huwd:
Adds a Lambda that:
- Reads in an SQS event Record, that should always have a TxMA event, and optionally has a service record
- Uses the TxMA data to either:
  - Take an existing service record, and iterate the `count_successful_logins` by one, and update the `last_accessed` to match the TxMA event timestamp OR
  - Add a new service record with a `count_successful_logins` of 1
- Finally write the output to a subsequent SQS queue

Note that the maps + filters mean that in the case we had an array of user services passed to us, and there was a duplicate service record (something that really _shouldn't_ be the case!) the code would take the first and ignore the others. I think it's probably best in that case we fail more loudly, so I've added in some validation to check for duplicate `client_ids` and throw if they are found. Usually I'd think that a bit OTT, but brought it in due to my current `.map()` obsession.

Replaces: https://github.com/alphagov/di-account-management-backend/pull/6


lambda/format-user-services/format-user-services.ts
```diff
@@ -1 +1,98 @@
...
+    QueueUrl: queueUrl,
+    MessageBody: JSON.stringify(messageBody),
+  };
+  const result = await client.send(new SendMessageCommand(message));
+  return result.MessageId;
+};
+
+export const handler = async (event: SQSEvent): Promise<void> => {
+  const { QUEUE_URL } = process.env;
+  if (!QUEUE_URL) {
```
huwd, line 79: I could not for the life of me get this bit to test correctly. Could use a second pair of 👀 to see what I'm doing wrong with jest?

lambda/format-user-services/format-user-services.ts
```diff
@@ -1 +1,98 @@
...
+    throw Error("No QUEUE_URL");
+  }
+  const { Records } = event;
+
+  const validatedRecords: SQSRecord[] = Records.map((record) =>
```
huwd, line 84: Anyone more familiar with spys?
I'd quite like to be able to add something to the handler function saying:
> it expects to call validatedRecords, once

etc etc...

lambda/format-user-services/format-user-services.ts
```diff
@@ -1 +1,98 @@
...
+  UserRecordEvent,
+  UserServices,
+  Service,
+  TxmaEventBody,
+} from "./models";
+
+const { AWS_REGION } = process.env;
+
+export const validateSQSRecord = (record: SQSRecord): SQSRecord => {
+  console.log(`Validating the SQS record body${record.body}`);
```
huwd, line 17: I should probably actually make this do something, woops.
Did we decide on a definition of valid?

lambda/format-user-services/format-user-services.ts
alex9smith: I'd argue that maybe you shouldn't - that sounds like you're testing for an implementation and not the behaviour. 

I think a clearer test would be to pass the handler a single invalid record and test that it never puts that record on the next queue.

huwd: Yeah fair enough, I guess it's small enough to spot if I've not brought in one of the key functions.
And Typescript / Linting doesn't let me have unused code now I think about it.

I'll get those validators sorted and pass something that doesn't get through them.

divyaparameswaran (approved):

PR #19, opened 2022-10-19
GUA-448: Allow lambdas to write logs to Cloudwatch

alex9smith:
In order to debug the system, we need to see the lambda logs and metrics. Give the built in [`AWSLambdaBasicExecutionRole`](https://console.aws.amazon.com/iam/home#policies/arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole) permissions to each of the new lambda functions, which is the [AWS recommended way](https://docs.aws.amazon.com/lambda/latest/dg/lambda-intro-execution-role.html#permissions-executionrole-features) to grant these permissions.

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -98,6 +98,8 @@ Resources:
   QueryUserServicesRole:
     Type: AWS::IAM::Role
     Properties:
+      ManagedPolicyArns:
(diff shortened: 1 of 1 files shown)


alex9smith:
I've included this commit in #20, so we don't need a separate PR

PR #20, opened 2022-10-19
GUA-448: Get the happy path working for lambda functions

alex9smith:
This PR fixes several small bugs to get the lambda functions all working together. We can now put an event on the input queue, see it processed by each lambda and end up in DynamoDB. 

The largest bug was that we weren't resolving the `Promise`s in the handler function for the first two lambdas, which meant the lambda stopped running before the messages got put on the queue to the next lambda. We were iterating over each record using `.forEach`, but that doesn't collect the returned objects and so discards the `Promise`s. We've now switched to `Records.map` which returns an array of `Promise`s, which we then wait to resolve using `await Promise.all( ... )`. 

We may want to tweak this behaviour in the future (perhaps with [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) ) to better allow us to do error catching per event, but we'll tackle that later - this PR is only about setting up the happy path.

---

It's probably easiest to review this commit by commit.  

This stack is deployed to the `dev` AWS account, we can test it by putting an event on the input queue and looking at the lambda logs in Cloudwatch and the DynamoDB table:

```bash
gds aws di-account-dev -- aws sqs send-message \
  --queue-url https://sqs.eu-west-2.amazonaws.com/985326104449/account-management-backend-TxMAInputDummyQueue-dKQby3vNf8m7 \
  --message-body '{"event_name":"event-name","timestamp":1666169856,"client_id":"client-id","component_id":"component-id","user":{"user_id":"user_id"}}'
```

If you change the user ID in the message body then look at the [summary of DynamoDB](https://eu-west-2.console.aws.amazon.com/dynamodbv2/home?region=eu-west-2#table?initialTagKey=&name=user_services&tab=overview) you can see more entries being added to the table.

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -246,7 +246,7 @@ Resources:
       DeadLetterQueue:
         Type: SQS
         TargetArn: !GetAtt WriteServicesDeadLetterQueue.Arn
-      Handler: store-service-record.handler
+      Handler: write-user-services.handler
       PackageType: Zip
       MemorySize: 128
       Runtime: nodejs16.x
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -100,6 +100,7 @@ Resources:
     Properties:
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+        - !Ref QueryUserServicesPolicy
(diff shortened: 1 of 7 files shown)


divyaparameswaran (approved):
LGTM

PR #21, opened 2022-10-19
BAU: Add initial README and licence

alex9smith:
We can add to the README as we work more things out - for example end-to-end tests or running a Lambda locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,28 @@
+# di-account-management-backend
+
+The backend and data store that serves the account management application.
+
+This is a serverless application for AWS. It's built and deployed using the [SAM CLI](https://aws.amazon.com/serverless/sam/).
+
+## Testing
+
+Each Lambda function is a separate NPM application and has its own unit tests.
+
+To run the tests for the `query-user-services` Lambda:
+
+```bash
+cd lambda/query-user-services
+npm ci
+npm run lint
+npm run test
+```
+
+## Deploying the application
+
+Deploy the application to the `dev` AWS account by running
+
+```bash
+cd infrastructure
+sam build
+gds aws di-account-dev -- sam deploy
+```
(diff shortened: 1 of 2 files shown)


divyaparameswaran (approved):

README.md
```diff
@@ -0,0 +1,28 @@
+# di-account-management-backend
+
+The backend and data store that serves the account management application.
+
+This is a serverless application for AWS. It's built and deployed using the [SAM CLI](https://aws.amazon.com/serverless/sam/).
```
divyaparameswaran, line 5: Should we specify the Node version here ?

huwd: That's a good point - I'm not sure where tells you that we're using node 16 currently... I think we do need that somewhere, But perhaps not here... if we need to keep it up to date I reckon we'd forget to change the README.

Perhaps we should look at adding `.nvmrc`'s to each lambda repo ([link here](https://github.com/nvm-sh/nvm#nvmrc)) or is there something we can specify in package.json?

Then perhaps we could have a per-requsites heading here about things you'd need to install to get that running.
I'd suggest we do anything like that in a followup PR though.

huwd (commented):
Looks good to me,
We should resolve @divyaparameswaran's comment first, then happy to ✅

README.md
alex9smith: Good point. You'd have to look at the Cloudformation template to see what version of Node we're deploying with. Not ideal...

We recommend using `nvm` in the new onboarding guide, so I think it makes sense to use the tools for that (`.nvmrc` files). That would mean we can just run 

```bash
nvm install
nvm use
```

in a lambda folder and get the right version. Much nicer than putting a directive in `package.json` which will tell you you've got the wrong version but not give you a way to install the correct one.

I'll do a follow up PR to add those in and put some instructions in the README

PR #22, opened 2022-10-20
BAU: Set Node.js version

alex9smith:
Set the Node.js version in an `.nvmrc` file so we can run `nvm install` and get the correct version of Node for local development, and add some instructions to the README.

Also make sure we're using that version of Node in the Github actions that run unit tests.


.nvmrc
```diff
@@ -0,0 +1 @@
+v16.18.0
```
huwd, line 1: Super interesting that you can put one up here, I'd assumed we'd need three. This is much better!

Can I check, if for some obscure reason we _needed_ to run a lambda in node 14, I'm assuming I _can_ put one of these next to a package.json and nvm will take the closest?

Was trying to work out these rules for eslintrc too, it does something similar

huwd (approved):
Looks good, non blocking but would be good to check on if we can ever override this?
Perhaps merge this and I'll drop a different `nvmrc` onto a test branch and see what happens

.nvmrc
alex9smith: Yup you're exactly right. NVM looks in the current directory, then if there isn't an `.nvmrc` file present it looks in the parent, and repeats until it gets to the top of the file system (where I assume it'll default to `latest`)

So we can put an `.nvmrc` file with a different version in one of the lambda folders and it'll use that first.

https://github.com/nvm-sh/nvm#nvmrc
> `nvm use` et. al. will traverse directory structure upwards from the current directory looking for the `.nvmrc` file. In other words, running `nvm use` et. al. in any subdirectory of a directory with an `.nvmrc` will result in that `.nvmrc` being utilized.

huwd: 🎉

PR #23, opened 2022-10-20
GUA - 431 dead letter queues

divyaparameswaran:
This PR adds the Dead Letter queue configuration to the format user service and query user service lambda. 
- If an error of any kind occurs during the process the record is sent to the Dead Letter Queue.
- I have also included some fix for linting errors in `format-user-services.test.ts` regarding an unused var.
- Also renamed format-user-record.test to format-user-service to match the handler ts.

### DIFF
--- a/lambda/format-user-services/tests/format-user-services.test.ts
+++ b/lambda/format-user-services/tests/format-user-services.test.ts
@@ -18,7 +18,7 @@ import {
   makeServiceRecord,
   makeSQSInputFixture,
 } from "./testHelpers";
-import { UserServices, Service } from "../models";
+import { UserServices, Service, TxmaEvent } from "../models";
 
 const sqsMock = mockClient(SQSClient);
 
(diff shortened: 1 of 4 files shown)


lambda/query-user-services/query-user-services.ts
```diff
@@ -71,7 +71,7 @@ const createUserRecordEvent = (
 };
 
 export const sendSqsMessage = async (
-  messageBody: UserRecordEvent,
+  messageBody: object,
```
huwd, line 74: Interesting, what caused the shift from a specific interface to `object`? I'm assuming `object` is just a generic for any `{}`.

Ideally I'd like to keep these specific?

lambda/query-user-services/tests/query-user-services.test.ts
```diff
@@ -290,3 +290,30 @@ describe("handler", () => {
     });
   });
 });
+
+describe("handler error handing ", () => {
+  let consoleErrorMock: jest.SpyInstance;
+
+  beforeEach(() => {
+    consoleErrorMock = jest.spyOn(global.console, "error").mockImplementation();
```
huwd, line 298: Ooh cool, I was wondering how to do that. Nice one.

lambda/format-user-services/tests/format-user-services.test.ts
```diff
@@ -18,7 +18,7 @@ import {
   makeServiceRecord,
   makeSQSInputFixture,
 } from "./testHelpers";
-import { UserServices, Service } from "../models";
+import { UserServices, Service, TxmaEvent } from "../models";
```
huwd: looks like there's a linting error here.
Let's have a chat about getting prettier set up in vscode it should autoremove things like this under some conditions!

huwd: ah wait, I see there's a clash with types declared below 🤔 
Ok perhaps best that this line becomes:

```suggestion
import type { UserServices, Service, TxmaEvent } from "../models";
```

Then remove TxmaEvent if it's not being used
and lines 144 and 161 should be `txmaEvent` to follow converntion as constants I think!

lambda/format-user-services/tests/format-user-services.test.ts
```diff
@@ -309,3 +309,50 @@ describe("handler", () => {
     });
   });
 });
+
+describe("handler error handling ", () => {
+  const messageID = "MyMessageId";
+  const sqsQueueName = "ToWriteSQS";
+  const queueURL = "http://my_queue_url";
+  const userId = "userID1234";
```
huwd: might need to go if it's not being used

lambda/format-user-services/tests/format-user-services.test.ts
```diff
@@ -309,3 +309,50 @@ describe("handler", () => {
...
+  const serviceClientID = "clientID1234";
```
huwd: might need to go if it's not being used

huwd (changes requested):
The linter has one or two things that it wants corrected.
Otherwise looks good

alex9smith (approved):
👍🏻

huwd (approved):

PR #26, opened 2022-10-21
GUA-432: Trigger alarms when the DLQs receive a message

alex9smith:
Set up a Cloudwatch alarm for each of the dead letter queues that's triggered when there's at least one message unprocessed in the queue.

The `ApproximateNumberOfMessagesVisible` metric seemed the most appropriate of those available [[1]. It's an approximation, but it'll always be greater than 0 if there's at least one message which is fine for this use case where we don't care how many messages there are.

There was also `NumberOfMessagesSent` which is "The number of messages added to a queue" but I wasn't sure if that would also be triggered if the whole Lambda failed to run and AWS put the message onto the DLQ, rather than us calling `sendMessage` from inside the lambdas.

For now these alarms don't do anything - we'll tackle sending notifications to Slack in future PRs.

---
I've deployed these alarms to the `dev` AWS account. You can see [an example here](https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#alarmsV2:alarm/account-management-backend-QueryDeadLetterQueueAlarm-IQ8IGKC99MIU?) where I put an event on a DLQ, the alarm triggered, then I purged the queue and the alarm reset.

[1]: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-available-cloudwatch-metrics.html

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -94,6 +94,8 @@ Resources:
     Type: AWS::SQS::Queue
     DeletionPolicy: Delete
     UpdateReplacePolicy: Delete
+    Properties:
+      MessageRetentionPeriod: 1209600
 
   QueryUserServicesRole:
     Type: AWS::IAM::Role
@@ -199,6 +201,8 @@ Resources:
     Type: AWS::SQS::Queue
     DeletionPolicy: Delete
     UpdateReplacePolicy: Delete
+    Properties:
+      MessageRetentionPeriod: 1209600
 
   FormatUserServicesRole:
     Type: AWS::IAM::Role
@@ -295,6 +299,8 @@ Resources:
     Type: AWS::SQS::Queue
     DeletionPolicy: Delete
     UpdateReplacePolicy: Delete
+    Properties:
+      MessageRetentionPeriod: 1209600
 
   WriteServiceRecordRole:
     Type: AWS::IAM::Role
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -146,6 +146,20 @@ Resources:
     DeletionPolicy: Delete
     UpdateReplacePolicy: Delete
 
+  QueryDeadLetterQueueAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      Namespace: "AWS/SQS"
+      MetricName: "ApproximateNumberOfMessagesVisible"
+      Dimensions:
+        - Name: "QueueName"
+          Value: !GetAtt QueryUserServicesDeadLetterQueue.QueueName
+      Statistic: "Sum"
+      Period: 300
+      EvaluationPeriods: 1
+      Threshold: 1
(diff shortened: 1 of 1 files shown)


infrastructure/template.yaml
```diff
@@ -94,6 +94,8 @@ Resources:
     Type: AWS::SQS::Queue
     DeletionPolicy: Delete
     UpdateReplacePolicy: Delete
+    Properties:
+      MessageRetentionPeriod: 1209600
```
huwd, line 98: That's 14 days right?

So is that: It gives us 14 days to respond to an error and read off the data to fix things up?
Sounds like a reasonable window given we'll probably want to act much sooner if anything ever crops up here.

infrastructure/template.yaml
```diff
@@ -146,6 +148,20 @@ Resources:
...
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      Namespace: "AWS/SQS"
+      MetricName: "ApproximateNumberOfMessagesVisible"
+      Dimensions:
+        - Name: "QueueName"
+          Value: !GetAtt QueryUserServicesDeadLetterQueue.QueueName
+      Statistic: "Sum"
+      Period: 300
+      EvaluationPeriods: 1
```
huwd, line 161: Right, so these are measured in intervals then?
And we saying trigger... if there is greater than 1 message on the queue, for _one_ period of 300 seconds?

I'm guessing this could be something we could usefully apply to the SQS queues between the lambda then?
Like if things back up so we have... (arbitrarily) 100 messages on the queue for longer than 300 seconds, then we alert saying "oi your scaling rules need a look, or something is broken?"

huwd (approved):

infrastructure/template.yaml
alex9smith: Yeah exactly, see my commit message on [4ff34fb](https://github.com/alphagov/di-account-management-backend/pull/26/commits/4ff34fb9c7cb19dd9bc7066b00141d5c97bf0416) for my thinking there.

infrastructure/template.yaml
alex9smith: >Right, so these are measured in intervals then?

It's a bit complicated but essentially yes. It's designed so you can do some clever analytics for your alarms.  For these ones it just checks every 5 minutes to see if there are any messages in the queue which is pretty simple. 

We can do more complicated things like setting the `Period` to 60, `EvaluationPeriods` to 5 and `DatapointsToAlarm` to 3 (not set here - it defaults to what you set `EvaluationPeriods` to). That would mean 'trigger the alarm if we go over the threshold number for at least 3 out of the last 5 minutes in a rolling window'.

The [AWS docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarm-evaluation) do a good job of explaining it.

PR #28, opened 2022-10-24
GUA-432: Send a message to Slack when an alarm is triggered 

alex9smith:
Set up AWS Chatbot to send messages to our Slack channel, mostly following what the dev platform pipeline [[1]] already does. Then add actions to each of the alarms to send a message to the new SNS topic when they're triggered.

This puts a message in Slack when an alarm goes off - see an example [here](https://gds.slack.com/archives/C011Y5SAY3U/p1666615861541789). Since that test message we've added the environment name to the alarm name so it's obvious which environment the alarm is for.

This Slack message topic and Chatbot config are prime candidates for moving to the central infrastructure stacks the Bravo team are writing. We may even not need the Chatbot config as one also gets created for the pipeline. We can consider that refactor when the central templates are ready.

[1]: https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3062726680/How+to+prepare+AWS+accounts+for+containing+SAM+deployment+pipelines

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -167,6 +167,7 @@ Resources:
   QueryDeadLetterQueueAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
+      AlarmName: !Join [ "-", [ !Ref AWS::StackName, !Ref Environment, QueryDeadLetterQueueAlarm ] ]
       Namespace: "AWS/SQS"
       MetricName: "ApproximateNumberOfMessagesVisible"
       Dimensions:
@@ -177,6 +178,9 @@ Resources:
       EvaluationPeriods: 1
       Threshold: 1
       ComparisonOperator: "GreaterThanOrEqualToThreshold"
+      AlarmActions:
+        - !Ref AlarmNotificationTopic
+      ActionsEnabled: true
 
   #######################
   # Formatting the Record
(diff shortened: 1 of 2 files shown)


infrastructure/template.yaml
```diff
@@ -16,6 +16,22 @@ Parameters:
       - "integration"
       - "production"
     ConstraintDescription: must be dev, build, staging, integration or production
+  SlackWorkspaceId:
+    Description: >
+      The ID of the Slack workspace where build notification messages are
+      posted. This is retrieved from the AWS Chatbot integration.
+    Type: String
+    Default: "T8GT9416G"
```
huwd, line 24: I take it this is the ID for the GDS workspace (gds.slack.com?)? And then `SlackChannelId` is our team tech channel?

infrastructure/template.yaml
```diff
@@ -177,6 +178,9 @@ Resources:
       EvaluationPeriods: 1
       Threshold: 1
       ComparisonOperator: "GreaterThanOrEqualToThreshold"
+      AlarmActions:
+        - !Ref AlarmNotificationTopic
+      ActionsEnabled: true
```
huwd, line 183: Cool, does this mean we could set up a bunch of these on all envs, and then do one additional PR to switch them "on" when we're ready for prod? Avoiding any dev noise in the meantime?

huwd (approved):

infrastructure/template.yaml
alex9smith: Yes that's right. I got these values from the dev platform pipeline stack in the `build` AWS account

alex9smith: In fact, this whole parameter definition is taken directly from that stack

infrastructure/template.yaml
alex9smith: Yeah I think that must be what this flag is designed for

PR #30, opened 2022-10-26
Add Log Group for Aws Serverless Functions

divyaparameswaran:
If a Log Group is not declared within CloudFormation, then this Log Group will be automatically created at run time and it won’t be managed by CloudFormation. The consequences are that the:

- Log Group will not be removed when the Stack is removed.
- Log Group attributes (such as retention) won’t be managed "as code" within CloudFormation.

These automatically created Log Groups will remain forever which can have huge impact on costs and security (the data stay forever while this is unexpected from company or regulatory rules).

A good practice is to declare the target Log Group with CloudFormation, matching the name that the AWS will use so that the Log Group resource is managed as code.

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -74,6 +74,8 @@ Resources:
   # Querying the Record
   ######################
   QueryUserServicesFunction:
+    DependsOn:
+      - QueryUserServicesFunctionLogGroup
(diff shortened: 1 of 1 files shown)


huwd (approved):
Great work, nice one spotting that and making proactive improvements.
Very happy to have sonarcloud appeased :)

PR #31, opened 2022-10-27
GUA-478 - Create ADR record in this repo

huwd:
Aim to do three things here:
- Explain why I've done this 
- Point to our key dependency on the TxMA ADR
- Move how we've chosen to persist this into this repo


docs/adr/0003-recording-service-usage-in-account.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Decision
+
+As a result of prior ADRs we have decided to:
+
+- Use DynamoDB as our data store, encrypted at rest with a non-default KMS key
+- Write small, single responsibility lambdas written in Typescript with minimal production dependencies
+- Separate of lambdas with SQS Queues to allow for async processing and scaling
+- Use dead letter queues to monitor and capture data from handled errors.
+
```
alex9smith, line 37:
```suggestion

We will store a single record per user and update that record as the users logs in to more services. This will make querying for a user's data very fast which is important, as the query will be blocking the synchronous user-facing HTTP request to load the service dashboard.

This means the time between us receiving an event from TxMA and writing it to the data store will be slightly longer since we're doing the data processing upfront. That's ok because data ingestion is an asynchronous, non-user facing process. Based on testing we expect it to take under 1 second in the worst case scenario when each lambda has to cold start which is an acceptable latency.

```
I think we should add this section to explain why we chose this specific architecture. We could have gone with a model that stores the incoming events as is (and actually I think we should still store them - will raise a separate ADR to discuss) but we decided not to because that would increase query time in the user facing part of the system.

alex9smith (approved):
Looks good! I've made some grammar fixes and also added a suggestion for some extra information to put in the persisting data ADR which I should have thought of on the last one.

PR #32, opened 2022-10-28
GUA-408: Add ADR to save unmodified event data

alex9smith:
This came to mind when we were talking to Paul J about our data model, and it also offers a solution to the problem that came up in pre-planning of potentially needing to delay showing future features to users while we collect 'enough' events to make the feature useful.

I've done this as an ADR that I'd like feedback and a decision on in the review step on Github. We can then either merge this PR (and accept the decision) or close it (and add a new ADR explaining why we didn't do this). Either is fine, but I think this is an important thing to make a decision on at this point.

It could possibly also have been an RFC; to be honest I'm not sure I understand where to use those. The end product of this decision would be an ADR though, so this felt correct.


huwd:
Yeah interesting one this, I see the replay problem and you've described it well.

I think there's a couple things for us to flesh out in comments here.

## Risk as impact / likelihood?
Am I right in thinking this boils down to: How much do we trust our current pipelines of SQS queues / dead letter queues?

The problem here seems to be that:
- once we've read of TxMA it's gone, with no replay-ability
- however there should only be two possible states for data here once it has been "read":
   1. It ends up in dynamo
   2. It ends up on a Dead Letter Queue and we have 14 days to address it

So in terms of likelihood, for us to lose data here we're looking at:
- An amazon outage
- A misconfiguration in our infrastructure (like I shipped the wrong SQS queue as an env var?)
- A misconfiguration in our code (like something isn't writing to DLQs properly)
- A failure in our test coverage
- A failure to address something within 14 days

Those are all things I'd quite like to have more confidence in 🤔 and decent mitigations for.
How much of this change is motivated by a lack of certainty here?

In terms of impact, I can see two scenarios:
- A user fails to have their service count incremented by 1 (I'm not sure a user would notice this?)
- A user fails to have a service they've used once appear on their dashboard (more serious but fixable)

So I'm thinking risks here are kind of low?

## Pros / Cons of modifying data

I'm less sure about this part. Trying to think through if the replayability is worth it in relation to other factors.

Pro of modifying the data:
We can just read a data structure out of the dynamo and pass it across to the frontend, knowing it's already ready to go.

Could we could also achieve this _and_ pass the TxMA event across, perhaps as a sub key/value? or does that defeat the point?

Pro of not storing all the data:
We have more certainty about what's in our store, if TxMA add additional superfluous fields, we can be sure they're not either bulking out our data store? Or that we've not ended up with PII that crosses any risk boundaries?

## Other options

If we wanted replay-ability, how long would we want it for?
I'm thinking what if we stuck another dynamo into this process, early on, perhaps adjacent / connected to the Query lambda. And that just sucks in an exact copy of the TxMA event. Stores it, and has a 30 day retention window.

If we've not replayed within 30 days, we dump it?

Does that get us the best of both world here or do we think that's needless duplication?

alex9smith:
> ## Risk as impact / likelihood?

Those are all fair points for the 'what if we lose an event' problem, and I'm pretty convinced by your argument that this is low impact to users and unlikely. Saving the results for 30 days is a good practical solution that probably wouldn't take much effort.

While I was writing this ADR the possibility to backfill data for a new feature was the more compelling reason to store the data for longer. Now I think about it a bit more, for that to be useful the feature would need to rely on only the event types we're already collecting (or a subset of them). For example, if we're only collecting login events this wouldn't help backfill a data store of when a user last proved their identity.

I think that makes it _less useful_, but still useful. We don't know all the features we want to use this data for yet, so it's hard to know how often we'd be able to backfill a new data store. I'd argue that it would still be worth storing the data even if we only backfilled once from it.

> ## Pros / Cons of modifying data
Perhaps I wasn't clear enough - the ADR is proposing to store this data _as well as_ modifying and storing data for each feature. So I don't think we need to worry too much about the pros and cons of how we store data because we can store it in the most appropriate format for each feature. 

On the risk of PII, this is another store of personal data so there is some inherent risk there. However, I think we should be able to trust the payloads we're getting from TxMA and if we / TxMA do detect something wrong we can run operations across the whole data store to drop the records or fields we shouldn't have. Perhaps I'm being a bit naive here though.

huwd:
Yeah Ok cool, storing it _as well_ makes sense.

I reckon if we do this, we bolt a new lambda and a new dynamo on to the front of our existing infra.

So a _new_ lambda reads off the TxMA input queue, it writes the payload to a separate Dynamo (with it's own retention - 30 days if we can't think of better sound OK?), and then passes the event on to a follow up queue, which is the start of our existing journey and the Query Lambda.

Reason I think we should do that, and not have the existing payload run through the entire chain of lambdas is:

- I'm feeling like I like how clean our existing dynamo data structure is (admittedly, bad reason...)
- If the risk if we'll need to replay an event because they're not making it through our lambdas, then I don't think we should be relying on those same lambdas to pass on the full event payload? I thought initially we were suggesting removing the three lambdas and writing directly to a database, moving the formatting to the "read" side. But if we're doing both, then if the formatting lambda failed we'd still lose the "full" TxMA payload, right?
- Also I like the idea of a separate retention period for the data, the unrefined stuff is just for replays. We can vary that based on our confidence in monitoring + ability to get to things quickly. The data store should be for "as long as the user needs it"?

huwd:
Assuming we were agreed on that, my proposal is:
- We do the bolt on ahead of our queue (should be pretty easy!)
- We modify this PR (so this conversation remains captured on it)
- Move what you've already got into "options considered"
- Document any final architecture in this ADR?
- Merge it up, and I'll make tickets for the work?

darrenhutton:
Makes sense.

Could make use of [DynamoDB streams](https://govukverify.atlassian.net/l/cp/GBZaAurS) here, rather than have a function that that stores and then passes to another queue

alex9smith:
> I reckon if we do this, we bolt a new lambda and a new dynamo on to the front of our existing infra.

I think we should use SNS to fan out the events to use multiple SQS queues and keep these systems completely separate. We can subscribe multiple SQS queues to an SNS topic and they'll [all get the same message when a message is put on the SNS topic](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-configure-subscribe-queue-sns-topic.html).

TxMA fan out

This sets us up for the future when we add a new feature that wants to consume a different subset of the events. We can subscribe a new SQS queue to the SNS topic, add a Lambda on the end of that to filter for just the events we want and then build whatever we want afterwards. 

Totally agree with all your reasons as to why to keep this separate. 

I'll add in a clarification that this is in addition to the other data store, this architecture plan and then tag you for a re-review.

darrenhutton:
I don't think this addresses the concern in the first comment about having to get enough events before displaying something to the user - but I don't understand that issue. Unless we think 30 days would be enough of a buffer of events

alex9smith:
I think the most sensible way of using DynamoDB streams here would be to write each event that comes from TxMA straight to a DynamoDB table, then use the stream from that table to trigger the rest of our feature pipelines. It'd look like this: 

image

However, we can [only subscribe two lambdas to a DynamoDB stream](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.Lambda.html) so at some point we'd need to recreate that SNS fan-out system from my comment above (or something equivalent) to provide messages for other downstream lambdas. So I think it makes sense at this point to put that fan-out system first and have the unmodified event store be just another subscriber. 

Does that make sense @huwd?

huwd:
That's very cool! and yeah that makes sense to me.

alex9smith:
Great, I'll update the ADR with this architecture and ask for a re-review

docs/adr/0004-store-all-event-data.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Architecture
+
+We will insert a new compenent before the existing pipeline to process a user's services.
+There will be a new lambda function which writes the unmodified event from TxMA to a new DynamoDB table.
+We'll then subscribe the lambda function at the start of the user services pipeline to the DynamoDB Stream for this new table.
+
+![Diagram showing events coming from TxMA being written to DynamoDB. The lambdas which ingest data for feature pipelines are triggered by the DynamoDB stream of that table.](images/adr-0004-architecture.png)
+
+We can only subscribe two lambda functions to a DynamoDB stream.
```
huwd, line 22:
```suggestion
We can [only subscribe two lambda functions to a DynamoDB stream](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.Lambda.html).
```

huwd: That link was useful from the comments

huwd (approved):
Looks good, one minor suggestion - take it or leave it.
@darrenhutton can I be a pain and ask for your ✅ here before we merge this?

Would be good to get the Architect's 🦭 of approval!

huwd (approved):
Nice one, great work.
I'll ticket this up on the backlog today

darrenhutton (approved):
This could be the start of something pretty big

PR #33, opened 2022-10-28
GUA-471: ADR to delete data from account

huwd:
Create ADR for removing data, WIP whilst we work out implications of #32 and check in with @darrenhutton on cross program approach?


darrenhutton (approved):
This is a good basis for learning to feed into [SSE implementation](https://github.com/alphagov/digital-identity-architecture/pull/205)

We will need to ensure other stores within the service can subscribe to these events pretty soon - thinking the IPV data store initially, poss TxMA. Perhaps raise at the weekly Tech Lead's meeting?

alex9smith (approved):
Nice! Do you think we should include an event schema here, or should that wait until this gets agreed on in a cross-programme ADR?

huwd:
🤔 good question @alex9smith, I suspect ours will just be:

```typescript
type UrnFdnSub = string;

export interface User {
  user_id: UrnFdnSub;
}
```

If we start this week, it may have to be different if others have an interest. Depends a bit on where folks are on pairwise I think?

I was tempted to go back and edit this, but I think let's chat about it with Tech leads first.
If there's massive interest I'll agree a schema with them and do a follow up ADR.

if not... I might sneakily put in a pull request editing this one and sneaking the schema we want as a team at the end of this one?

PR #34, opened 2022-11-02
BAU: Update prerequisites and testing instructions

alex9smith:
Adds `esbuild` as a dependency since this is required for `sam build`.

Also add instructions to send a test event to the application's input queue.


alex9smith:
The failing deploy action shouldn't have run (that's fixed in #35), but it fails when assuming the deploy role because AWS restricts that to the `main` branch and not because deploys are broken.

divyaparameswaran (approved):

PR #36, opened 2022-11-02
GUA-467 Cloudformation for delete stack

huwd:
- Add cloud formation for delete stack components as defined in [0005-deleting-service-usage-in-account.md](https://github.com/alphagov/di-account-management-backend/blob/main/docs/adr/0005-deleting-service-usage-in-account.md)
- Add placeholder basic files for `delete-user-services` so `sam build` and `sam deploy` works

The lambdas do need to work, as there's a [follow up ticket to write and test them properly](https://govukverify.atlassian.net/jira/software/c/projects/GUA/boards/195?modal=detail&selectedIssue=GUA-469).
What we do need is to get enough basic components in place so we can `sam deploy` successfully.

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -646,20 +646,14 @@ Resources:
                 - Arn: !GetAtt UserServicesStore.Arn
           - Effect: Allow
             Action:
-              - kms:*
-            Resource: !GetAtt QueueKmsKey.Arn
-          - Effect: Allow
-            Action:
-              - kms:*
-            Resource: !GetAtt LoggingKmsKey.Arn
-          - Effect: Allow
-            Action:
-              - kms:*
-            Resource: !GetAtt LambdaKMSKey.Arn
-          - Effect: Allow
-            Action:
-              - kms:*
-            Resource: !GetAtt DatabaseKmsKey.Arn
+              - kms:Decrypt
+              - kms:GenerateDataKey*
+              - kms:ReEncrypt*
+            Resource:
+              - !GetAtt QueueKmsKey.Arn
+              - !GetAtt DatabaseKmsKey.Arn
+              - !GetAtt LoggingKmsKey.Arn
+              - !GetAtt LambdaKMSKey.Arn
 
   DeleteUserServicesFunctionLogGroup:
     Type: AWS::Logs::LogGroup
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -555,6 +555,157 @@ Resources:
         - !Ref AlarmNotificationTopic
       ActionsEnabled: true
 
+  ######################
+  # Deleting the Record
+  ######################
+
+  UserAccountDeletionTopic:
+    Type: AWS::SNS::Topic
+    Properties:
+      DisplayName: UserAccountDeletionTopic
+      KmsMasterKeyId: !Ref SnsKmsKey
+      TopicName: UserAccountDeletion
+
+  DeleteUserServicesSubscription:
+    Type: AWS::SNS::Subscription
+    Properties:
+      Endpoint: !GetAtt DeleteUserServicesFunction.Arn
+      Protocol: lambda
+      RedrivePolicy:
+        deadLetterTargetArn: !GetAtt UserAccountDeletionTopicDeadLetterQueue.Arn
+      TopicArn: !Ref UserAccountDeletionTopic
+
+  DeleteUserServicesFunction:
+    DependsOn:
+      - DeleteUserServicesFunctionLogGroup
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: !Sub ${Environment}-${AWS::StackName}-delete-user-services
+      Architectures: ["x86_64"]
+      CodeUri: ../lambda/delete-user-services/
+      DeadLetterQueue:
+        Type: SQS
+        TargetArn: !GetAtt DeleteUserServicesDeadLetterQueue.Arn
+      Handler: delete-user-services.handler
+      KmsKeyArn: !GetAtt LambdaKMSKey.Arn
+      PackageType: Zip
+      MemorySize: 128
+      Runtime: nodejs16.x
+      Role: !GetAtt DeleteUserServicesRole.Arn
+      Timeout: 5
+      Environment:
+        Variables:
+          TABLE_NAME: !Ref UserServicesStoreTableName
+          DLQ_URL: !Ref DeleteUserServicesDeadLetterQueue
+    Metadata:
+      BuildMethod: esbuild
+      BuildProperties:
+        Minify: true
+        Target: "es2020"
+        Sourcemap: true
+        EntryPoints:
+          - delete-user-services.ts
+
+  DeleteUserServicesRole:
+    Type: AWS::IAM::Role
+    Properties:
+      PermissionsBoundary: !If
+        - UsePermissionsBoundary
+        - !Ref PermissionsBoundary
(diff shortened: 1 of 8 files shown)


huwd:
Ah so close on a deploy.
There's a checkov hook sad about the passthrough Lambda, will work out what's going on there.

infrastructure/template.yaml
```diff
@@ -517,6 +517,260 @@ Resources:
...
+          - Effect: Allow
+            Action:
+              - sqs:ReceiveMessage
+              - sqs:DeleteMessage
+              - sqs:GetQueueAttributes
+            Resource: !GetAtt UserDeleteAccountQueue.Arn
+          - Effect: Allow
+            Action:
+              - sns:Publish
+            Resource: !Ref UserAccountDeletionTopic
```
huwd, line 651: This is odd but couldn't get it to like a `!GetAtt UserAccountDelegationTopic.Arn`
Not really sure why.

huwd:
Problem is:
```
Hook failed with message: Checks failed. CKV_AWS_116: Ensure that AWS Lambda function is configured for a Dead Letter Queue(DLQ) (https://docs.bridgecrew.io/docs/ensure-that-aws-lambda-function-is-configured-for-a-dead-letter-queue-dlq)
```
Will sort that tomorrow.
Hadn't put a DLQ on this because I figured not much could go wrong, but... fair enough.

huwd:
Came back later and looks like it deployed!

huwd:
FYI, we've simplified this at @alex9smith's suggestion.
He rightly pointed out we may not need the passthrough lambda!

Our current assumption is there's no bottleneck for the fargate app to be writing directly to the SNS topic. It seems like a good assumption to me and has allowed us to simplify a bit.

So that's a change from here, I'll update the diagram:
https://github.com/alphagov/di-account-management-backend/blob/main/docs/adr/0005-deleting-service-usage-in-account.md

infrastructure/template.yaml
huwd: Alex solved that one, the docs show that yeah doesn't support ARN.
So this is just another strange cloudformation inconsistency:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sns-topic.html#aws-resource-sns-topic-return-values-fn--getatt

huwd:
Should be adjusted to reflect:
https://github.com/alphagov/di-account-management-backend/pull/39

infrastructure/template.yaml
```diff
@@ -517,6 +517,156 @@ Resources:
...
+          - Effect: Allow
+            Action:
+              - kms:*
+            Resource: !GetAtt LambdaKMSKey.Arn
+          - Effect: Allow
+            Action:
+              - kms:*
+            Resource: !GetAtt DatabaseKmsKey.Arn
+
+  DeleteUserServicesFunctionLogGroup:
```
alex9smith, line 625: Can you add the CSLS subscription for this log group like the ones added in #42? You might need to rebase to get the condition

infrastructure/template.yaml
```diff
@@ -517,6 +517,156 @@ Resources:
...
+            Resource: !GetAtt DeleteUserServicesDeadLetterQueue.Arn
+          - Effect: Allow
+            Action:
+              - dynamodb:DeleteItem
+            Resource:
+              - !GetAtt UserServicesStore.Arn
+              - !Sub
+                - "${Arn}/*"
+                - Arn: !GetAtt UserServicesStore.Arn
+          - Effect: Allow
```
alex9smith, line 608: These need to be the more restrictive KMS policies we switched to in #41

infrastructure/template.yaml
```diff
@@ -517,6 +517,156 @@ Resources:
...
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - sqs:SendMessage
+            Resource: !GetAtt DeleteUserServicesDeadLetterQueue.Arn
+          - Effect: Allow
+            Action:
+              - dynamodb:DeleteItem
```
alex9smith, line 602: Huh is this all it needs? I'd expect it to also need some sort of query permission. Happy to be proven wrong though

alex9smith (changes requested):
Looks great! Just a couple of changes needed for things we've added since you started working on this

infrastructure/template.yaml
huwd: Yeah I had a similar 🧠, I think what swayed me was the command that we're actually firing from the lambda, there's no follow up query, so unless it's under the hood, we'll see.

I'll get a test together to prove it.

infrastructure/template.yaml
huwd: yeah great catch, I'll do that now.

infrastructure/template.yaml
huwd: Done here:
https://github.com/alphagov/di-account-management-backend/pull/36/commits/c13c10d589911e0dfc0ce0961d82c867e3cc69e7

infrastructure/template.yaml
```diff
@@ -555,6 +555,165 @@ Resources:
...
+      RedrivePolicy:
+        deadLetterTargetArn: !GetAtt UserAccountDeletionTopicDeadLetterQueue.Arn
+      TopicArn: !Ref UserAccountDeletionTopic
+
+  DeleteUserServicesFunction:
+    DependsOn:
+      - DeleteUserServicesFunctionLogGroup
+    Type: AWS::Serverless::Function
+    Properties:
+      Architectures: ["x86_64"]
```
divyaparameswaran, line 583: Could we include a Function Name here for the log groups please  - https://github.com/alphagov/di-account-management-backend/blob/main/infrastructure/template.yaml#LL119C7-L119C19

infrastructure/template.yaml
```diff
@@ -555,6 +555,165 @@ Resources:
...
+  ######################
+
+  UserAccountDeletionTopic:
+    Type: AWS::SNS::Topic
+    Properties:
+      DisplayName: UserAccountDeletionTopic
+      KmsMasterKeyId: !Ref SnsKmsKey
+      TopicName: UserAccountDeletion
+
+  DeteUserServicesSubscription:
```
alex9smith, line 569:
```suggestion
  DeleteUserServicesSubscription:
```

divyaparameswaran (commented):
Looks good to me , but I'll wait for Alex to approve as he had previously requested changes .

infrastructure/template.yaml
huwd: yeah just spotted that too, forced pushed! thanks!

infrastructure/template.yaml
huwd: ah excellent! thanks was wondering why yours were named differently to mine!

huwd: Done:
https://github.com/alphagov/di-account-management-backend/pull/36/files#diff-1d6a0a19507d7cdddd8ff209f817d2d18f3fb22ceec6b0eb332b0a14292c69b0R583

alex9smith:
You've got a subscription for the lambda to the SNS topic, but I don't think that'll be enough as there's nothing here giving SNS permission to invoke the lambda. 

I think you'll either need to add[ an `Events` block to the lambda definition](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-events), or a separate `AWS::Lambda::Permission` resource

alex9smith (approved):
Nice one!

PR #38, opened 2022-11-04
GUA-469 delete lambda

huwd:
## How to test

1. Pre-populate a dynamoDB record with our target user_id
```bash
gds aws di-account-dev -- aws dynamodb put-item \
    --table-name user_services \
    --item '{ "user_id": { "S": "test-to-delete"}}'
```
Or, [create an item in the AWS console](https://eu-west-2.console.aws.amazon.com/dynamodbv2/home?region=eu-west-2#edit-item?table=user_services&route=ROUTE_ITEM_EXPLORER&itemMode=1) with `user_id` = "test-to-delete"

2. Query to check that your new record is there:
```bash
gds aws di-account-dev -- aws dynamodb get-item \
      --table-name user_services \
      --key '{"user_id": {"S": "test-to-delete"}}'
```

or to see all entries:
```bash
gds aws di-account-dev -- aws dynamodb scan --table-name user_services
```

You can also explore the [AWS console here](https://eu-west-2.console.aws.amazon.com/dynamodbv2/home?region=eu-west-2#item-explorer?initialTagKey=&table=user_services)

3. Get the ARN for your SNS topic

Either check in the console, [starting here](https://eu-west-2.console.aws.amazon.com/sns/v3/home?region=eu-west-2#/topics)

OR

```bash
 gds aws di-account-dev -- aws di-account-dev -- aws sns list-topics
```

4. Fire an SNS event

```
 gds aws di-account-dev -- aws sns publish \
    --topic-arn SNS_TOPIC_ARN_HERE \
    --message '{ "user_id": "test-to-delete" }'
```

5. Repeat step (2) to see if there's still a dynamoDB entry

### DIFF
--- a/lambda/delete-user-services/delete-user-services.ts
+++ b/lambda/delete-user-services/delete-user-services.ts
@@ -1,3 +1,64 @@
-const handler = async (): Promise<void> => console.log("Delete function");
+import { SQSEvent } from "aws-lambda";
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
+import {
+  DynamoDBDocumentClient,
+  DeleteCommand,
+  DeleteCommandOutput,
+} from "@aws-sdk/lib-dynamodb";
+import {
+  SendMessageCommand,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import { UserData } from "./models";
 
-export default handler;
+const { TABLE_NAME } = process.env;
+const marshallOptions = {
+  convertClassInstanceToMap: true,
+};
+const translateConfig = { marshallOptions };
+
+const dynamoClient = new DynamoDBClient({});
+const dynamoDocClient = DynamoDBDocumentClient.from(
+  dynamoClient,
+  translateConfig
+);
+
+const sqsClient = new SQSClient({});
+const { DLQ_URL } = process.env;
+
+export const validateUserData = (userData: UserData): UserData => {
+  if (userData.user_id) {
+    return userData;
+  }
+  throw new Error(`userData did not have a user_id ${userData}`);
+};
+
+export const deleteUserData = async (
+  userData: UserData
+): Promise<DeleteCommandOutput> => {
+  const command = new DeleteCommand({
+    TableName: TABLE_NAME,
+    Key: { user_id: userData.user_id },
+  });
+  return dynamoDocClient.send(command);
+};
+
+export const handler = async (event: SQSEvent): Promise<void> => {
+  await Promise.all(
+    event.Records.map(async (record) => {
+      try {
+        const userData: UserData = JSON.parse(record.body);
+        validateUserData(userData);
+        await deleteUserData(userData);
+      } catch (err) {
+        console.error(err);
+        const message: SendMessageRequest = {
+          QueueUrl: DLQ_URL,
+          MessageBody: record.body,
+        };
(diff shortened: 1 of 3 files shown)


huwd:
Needs adjusting as it'll now be invoked directly from an SNS topic not an SQS queue

lambda/delete-user-services/delete-user-services.ts
```diff
@@ -1,3 +1,64 @@
...
+} from "@aws-sdk/lib-dynamodb";
+import {
+  SendMessageCommand,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import { UserData } from "./models";
 
-export default handler;
+const { TABLE_NAME } = process.env;
```
alex9smith, line 15: We may as well combine this with 
```
const { DLQ_URL } = process.env;
```
on line 28 (I'm surprised there's not a ESLint rule to catch this).

lambda/delete-user-services/tests/delete-user-services.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    process.env.TABLE_NAME = "TABLE_NAME";
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test("deletes item from DynamoDB", async () => {
+    await deleteUserData(TEST_USER_DATA);
+    expect(dynamoMock.commandCalls(DeleteCommand).length).toEqual(1);
```
alex9smith, line 28: Could you add another assertion that it call the delete command with the user ID? 
We could use the [Jest matchers](https://github.com/m-radzikowski/aws-sdk-client-mock#jest-matchers) like this (untested)

```suggestion
    expect(dynamoMock.commandCalls(DeleteCommand).length).toEqual(1);
    expect(dynamoMock).toHaveReceivedCommand(DeleteCommand, {
      TableName: process.env.TABLE_NAME,
      Key: { user_id: TEST_USER_DATA.user_id },
    });
```

alex9smith (changes requested):
Just two things - one super nit-pick style change (take or leave it), and one extra test I'd like to see.

lambda/delete-user-services/delete-user-services.ts
huwd: Good spot, done!

lambda/delete-user-services/tests/delete-user-services.test.ts
huwd: Ok so this solves the problem:
https://github.com/alphagov/di-account-management-backend/pull/38/commits/d8f8e704ddf40339d7e02c3df2eb5a681db895da

Turns out when I console.logged the env it _was_ there. So not a jest thing.

Seems to work fine on the lambda, so I wonder if there's something about the way Jest Invokes these... like not the file just the fucntion?

Bit odd.

Anyway there's no re-use around the file for these, so seems safe to put them directly into the function.

huwd:
🤔 odd, another example of a flakey test here:
- [Job before](https://github.com/alphagov/di-account-management-backend/actions/runs/3445066252/jobs/5748350316)
- zero code change, [after re-running](https://github.com/alphagov/di-account-management-backend/actions/runs/3445066252/jobs/5748468073)

What on earth? How is there any state here to flake over?

huwd:
My best bets is this i something to do with that process.env stuff and the test order?

alex9smith (approved):

PR #39, opened 2022-11-04
BAU: Update Architectural Diagram to reflect new plan

huwd:
After starting work on [1] and [2] we realised the passthrough lambdas weren't really offering us much value, so we've removed them here.

The theory initially was that we'd want an SQS queue, and a scalable lambda allow us to efficently write to the SNS topic. We don't think that's true anymore. We should be able to write directly from the fargate containers, which themselves can scale horizontally, and we don't expect to hit a bottleneck here.

So let's remove a few components and simplify things a bit

[1]: https://github.com/alphagov/di-account-management-backend/pull/36
[2]: https://github.com/alphagov/di-account-management-backend/pull/38


alex9smith (commented):
We can simplify this further - we don't need the SQS queue between the SNS topic and the lambda either:

https://docs.aws.amazon.com/lambda/latest/dg/with-sns.html

darrenhutton:
Is it therefore:

1. Fargate app makes publish call to SNS
2. Delete Lambda function subscribes to the SNS topic
   - SNS topic [makes use of SQS DLQ](https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html) if delivery fails
4. Delete Lambda function does its thing
   - If Delete Lambda function fails, error handling config is set to use SQS DLQ

alex9smith (approved):

docs/adr/0005-deleting-service-usage-in-account.md
```diff
@@ -18,22 +18,23 @@ We must ensure that when a user deletes their account their service data is also
...
 
-We will set up a serverless infrastructure, triggered upon a user deleting their account.
+We will set up an SNS topic for a delete event, triggered upon a user deleting their account.
 
 The event will contain a user_id matching the desired record in the data store.
 
-We will write through a series of queues to a lambda which will issue a delete command for that record.
+Failure of the SNS queue to deliver an event to subscribers will invoke an [SNS Dead Letter queue](https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html).
+
+The Topic will subscribe to invoke a lambda which will issue a delete command for the relevant record.
```
alex9smith, line 27:
```suggestion
We will subscribe a lambda to the topic which will issue a delete command for the relevant record.
```

PR #40, opened 2022-11-07
Add function name to Lambda to create a log group 

divyaparameswaran:
Custom log group names are not allowed in AWS and they follow the pattern of /aws/lambda/{functionName} . Add function name to the lambdas to create the correct Log Group Name. 

Thanks to @alex9smith


alex9smith (approved):
👍🏻

PR #43, opened 2022-11-10
GUA-500 Add vpc and relevant roles to the lambda

divyaparameswaran:
- Use the `enhanced-vpc` generated in the frontend app 
- Configure the config on the lambda function .
- Give the lambda policy the correct access.


huwd (approved):

huwd:
Looks good, I'll rebase [this](https://github.com/alphagov/di-account-management-backend/pull/38) to do the same

PR #44, opened 2022-11-11
GUA-487: Create DynamoDB table for raw events

alex9smith:
This table has streams [[1]] enabled, which will become the input trigger for the `query-user-services` lambda. Items in this table will be written once and not updated, so I've used the `NEW_IMAGE` stream setting which will only include the newly written data in the stream event.

The table also has two keys - `id` and `timestamp`. DynamoDB requires a hash key that's used for the primary index. We don't want the lambda that writes these items to parse the TxMA event because that might fail, so we'll need to generate this ID in the lambda.

We're also using `timestamp` as a RANGE key. If we do ever need to use this table to replay events, it'll probably be over a time range. Using the timestamp as a key gives us a secondary index that'll speed up time-based queries.

We want to use Dynamo's TTL feature [[2]] to expire data in this table after a period of time. This requires a top-level attribute in the item that is:
1. a Number type
2. a timestamp in Unix epoch time format in seconds
3. not more than 5 years ago [[3]]

Dynamo will regularly scan the table and delete items that are older than the timestamp in this attribute. If the attribute's not present in an item, DynamoDB won't delete it. I've called this attribute `remove_at`, and the lambda writing these items will need to generate a timestamp in the future and include this attribute.

[1]: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html
[2]: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html
[3]: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/howitworks-ttl.html

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -6,6 +6,9 @@ Parameters:
   UserServicesStoreTableName:
     Type: String
     Default: user_services
+  RawEventsStoreTableName:
+    Type: String
+    Default: raw_events
   Environment:
     Description: "The environment type"
     Type: "String"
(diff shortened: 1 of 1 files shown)


infrastructure/template.yaml
```diff
@@ -65,7 +68,50 @@ Globals:
...
+        SSEType: KMS
+        KMSMasterKeyId: !GetAtt DatabaseKmsKey.Arn
+      StreamSpecification:
+        StreamViewType: NEW_IMAGE
+      TimeToLiveSpecification:
+        AttributeName: remove_at
+        Enabled: true
+      Tags:
+        - Key: Product
+          Value: GOV.UK Sign In
```
huwd, line 103: ooh well remembered on the tags, not sure I've been doing that consistently.

Let's work out what we want to tag, could be every resource right?
Or just sites of persistence?

huwd (approved):

infrastructure/template.yaml
alex9smith: We should steal what the Bravo team have done [here](https://github.com/alphagov/di-accounts-infra/blob/main/platform-common/template.yaml#L19-L34) and set the values as params. That'll remove a bit of repetition at least. 

I wonder if you can use the `Globals` thing as well to tag everything?

alex9smith: Ah no rubbish.

>The Globals section is supported by the AWS::Serverless::Function, AWS::Serverless::Api, AWS::Serverless::HttpApi, and AWS::Serverless::SimpleTable resources. [[1]]

[1]: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html

PR #45, opened 2022-11-11
GUA-487: Switch user services pipeline to Dynamo stream

alex9smith:
Switches the first lambda in the user services pipeline to use the DynamoDB table stream added in #44. This is part of the work to update the system architecture to reflect [ADR 0004](https://github.com/alphagov/di-account-management-backend/blob/main/docs/adr/0004-store-all-event-data.md).

This lambda is only subscribed to events when new items are written to the table. We're also enabling DynamoDB's TTL feature on this table, and we don't want the delete events from that to cause errors or (worse) be treated as new events.

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -166,10 +166,19 @@ Resources:
       Architectures: ["x86_64"]
       CodeUri: ../lambda/query-user-services/
       Events:
-        TxMAInputDummyQueue:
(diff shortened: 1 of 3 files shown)


huwd (approved):

huwd (approved):
Corrected IAMs + that syntax I missed (woops!)

PR #48, opened 2022-11-14
GUA-486 Save raw Events 

divyaparameswaran:
- This lambda reads from the SQS queue and saves the TXma event without modification to the raw_events table .
- Update the GitHub workflows to run the tests on merge .

### DIFF
--- a/lambda/save-raw-events/save-raw-events.ts
+++ b/lambda/save-raw-events/save-raw-events.ts
@@ -28,8 +28,7 @@ const sqsClient = new SQSClient({});
 const { DLQ_URL } = process.env;
 
 const getEventId = (): string => {
-  const array = new Uint16Array(10);
-  return `${Date.now()}-${crypto.getRandomValues(array)[0]}`;
+  return `${crypto.randomUUID}`;
 };
 
 const getTTLDate = (): number => {
--- a/lambda/save-raw-events/save-raw-events.ts
+++ b/lambda/save-raw-events/save-raw-events.ts
@@ -13,7 +13,6 @@ import {
 } from "@aws-sdk/client-sqs";
 import { TxmaEvent, UserData } from "./models";
 
-const { TABLE_NAME } = process.env;
 const marshallOptions = {
   convertClassInstanceToMap: true,
 };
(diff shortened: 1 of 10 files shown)


lambda/save-raw-events/save-raw-events.ts
```diff
@@ -0,0 +1,92 @@
...
+  PutCommandOutput,
+} from "@aws-sdk/lib-dynamodb";
+import {
+  SendMessageCommand,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import { TxmaEvent, UserData } from "./models";
+
+const { TABLE_NAME } = process.env;
```
huwd, line 16: These could be collapsed with L27.

But! I'd suggest moving the this line to L64 so it sits within the `writeRawTxmaEvent` function,
And L27 down to the handler function so it's on something like L76.

There's some occasional test flake that I was getting, ([see here](https://github.com/alphagov/di-account-management-backend/pull/46)), best I can tell it's something to do with the order the tests are run in, and the env occasionally being null.

I wondered if it's to do with the way that jest brings in a function vs a file?

In any case, these two aren't widely re-used so lets put them into the functions as a specific concern of that function

lambda/save-raw-events/save-raw-events.ts
```diff
@@ -0,0 +1,92 @@
...
+    txmaEvent.client_id !== undefined &&
+    txmaEvent.timestamp !== undefined &&
+    txmaEvent.event_name !== undefined &&
+    txmaEvent.component_id !== undefined &&
+    txmaEvent.user !== undefined
+  ) {
+    validateUser(txmaEvent.user);
+  } else {
+    throw new Error(`Could not validate UserServices ${txmaEvent}`);
+  }
```
huwd, line 58: I wonder if the existence of this here means we could lose some of this validation from the query step?
Or if once this is merged that's an opportunity to explore DRY-ing this out?

lambda/save-raw-events/tests/save-raw-events.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    consoleErrorMock.mockRestore();
+    jest.clearAllMocks();
+  });
+  test("logs the error message", async () => {
+    await handler(TEST_SQS_EVENT);
+    expect(consoleErrorMock).toHaveBeenCalledTimes(1);
+  });
+  test("sends the event to the dead letter queue", async () => {
+    await handler(TEST_SQS_EVENT);
+    expect(sqsMock.commandCalls(SendMessageCommand).length).toEqual(1);
```
huwd, line 227: Take a look and see if Alex's suggestion here makes sense for this test too?
https://github.com/alphagov/di-account-management-backend/pull/38#discussion_r1019990660

I thought the adjustment ended up with a better test in the end.
You can see my change here:
https://github.com/alphagov/di-account-management-backend/commit/24bd4947089d12ba04b9c3e2e99f662236e3f3ea
`lambda/delete-user-services/tests/delete-user-services.test.ts`

huwd (commented):
Looks good, couple of pretty minor suggestions from me.

You might also want to squash a few of these down into the previous commits.

lambda/save-raw-events/save-raw-events.ts
divyaparameswaran: Oh yes makes sense . I can do that after this is tested and all works fine, if that's alright with you .

huwd (approved):

huwd (approved):

PR #50, opened 2022-11-16
GUA- 485 Unmarshall db stream

divyaparameswaran:
- Use the dynamodb util unmarshall function to unmarshall the DynamoDbrecord to TxmaEvent.
- Update tests to reflect the changes.

### DIFF
--- a/lambda/query-user-services/query-user-services.ts
+++ b/lambda/query-user-services/query-user-services.ts
@@ -1,12 +1,12 @@
 import { DynamoDBStreamEvent } from "aws-lambda";
-import { DynamoDB } from "aws-sdk";
+import { unmarshall } from "@aws-sdk/util-dynamodb";
 import {
   SendMessageCommand,
   SendMessageRequest,
   SQSClient,
 } from "@aws-sdk/client-sqs";
 import { DynamoDBDocumentClient, GetCommand } from "@aws-sdk/lib-dynamodb";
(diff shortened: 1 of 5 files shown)


lambda/query-user-services/query-user-services.ts
```diff
@@ -90,7 +91,13 @@ export const handler = async (event: DynamoDBStreamEvent): Promise<void> => {
   await Promise.all(
     Records.map(async (record) => {
       try {
-        const eventBody = record.dynamodb?.NewImage?.event.S as string;
+        const eventBody = JSON.stringify(
```
alex9smith, line 94: There may be a good reason for this because DynamoDB is _weird_, but calling `JSON.stringify()` on an object and then `JSON.parse()` the line after doesn't look quite right. 

Could we do it like this instead? (I've checked the unit tests pass but not deployed it)
```typescript
      try {
        const txmaEvent = unmarshall(
          record.dynamodb?.NewImage?.event.M as {
            [key: string]: AttributeValue;
          }
        ) as TxmaEvent;
        validateTxmaEventBody(txmaEvent);
```

divyaparameswaran: hmm yeah .. honestly I didn't think about this as I was going back and forth with different changes to get this working. I will deploy once and see if the lambdas are happy

alex9smith (approved):

PR #54, opened 2022-11-21
GUA-514: Log record on error

alex9smith:
To aid with debugging the integration with TxMA.

We'll want to revert this before we deploy the backend to production to make sure we don't log any personal or sensitive data.


alex9smith:
Closing this because we figured out the problem before needing to log the events to CloudWatch

PR #56, opened 2022-11-28
GUA-470: Add the SNS topic ARN as a SSM parameter

huwd:
We'll need to be able to get this ARN into our frontend stack, but as the cloudformation lives in separate template files, the way we're going to exchange it is to have this deploy write the ARN to an SSM parameter, which we can then call out of the frontend stack.

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -632,6 +632,14 @@ Resources:
       KmsMasterKeyId: !Ref SnsKmsKey
       TopicName: UserAccountDeletion
 
+  UserAccountDeletionTopicArnParameter:
+    Type: AWS::SSM::Parameter
+    Properties:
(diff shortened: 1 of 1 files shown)


alex9smith (approved):

infrastructure/template.yaml
```diff
@@ -632,6 +632,14 @@ Resources:
       KmsMasterKeyId: !Ref SnsKmsKey
       TopicName: UserAccountDeletion
 
+ UserAccountDeletionTopicArnParameter:
```
alex9smith, line 635: The linter's not happy - I think it's your indentation here

huwd: yeah just spotted that too, thanks, switching to the machine where I have `cfn-lint` installed 😭

PR #58, opened 2022-12-07
GUA-470: Stringify userData object so errors print something useful

huwd:
Something I caught whilst trying to debug the lambda.

If there's something wrong way the Message payload outputting [object object] is not useful.
Let's stingify the JSON.

I checked this agianst a number of other error states to check this can't cause an error in and of itself:
`JSON.stringify("")` -> `'""'`
`JSON.stringify({})` -> `'{}'`
`JSON.stringify(undefined)` -> ` `
`JSON.stringify(null)` -> `null`
`JSON.stringify({"bloop")` -> `│Uncaught SyntaxError: Unexpected token ')'` _does_ throw an error but shouldn't come up IMO?

Worth folks checking those in the review, JSON.stringify seemed pretty permissive to me!


huwd:
Urgh I need to get prettier set up on this machine. Forced push to overcome the linting woes.

divyaparameswaran (approved):

PR #59, opened 2022-12-07
GUA-470: Ship the SNS KMS Key ID as an SSM parameter

huwd:
We'll need to pick up on this in the frontend application

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -1188,3 +1188,11 @@ Resources:
     Properties:
       AliasName: !Sub "alias/${AWS::StackName}/${Environment}/SnsKmsKey"
       TargetKeyId: !Ref SnsKmsKey
+
+  SnsKmsKeyIDParameter:
+    Type: AWS::SSM::Parameter
+    Properties:
+      Description: The ID of the SNS KMS Key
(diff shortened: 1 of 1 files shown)


infrastructure/template.yaml
```diff
@@ -1188,3 +1188,11 @@ Resources:
     Properties:
       AliasName: !Sub "alias/${AWS::StackName}/${Environment}/SnsKmsKey"
       TargetKeyId: !Ref SnsKmsKey
+
+  SnsKmsKeyARNParameter:
```
domantas-gds, line 1192: Would update ARN to ID

infrastructure/template.yaml
```diff
@@ -1188,3 +1188,11 @@ Resources:
...
+    Type: AWS::SSM::Parameter
+    Properties:
+      Description: The ARN of the SNS KMS Key
```
domantas-gds, line 1195: And ARN to ID in description :)

domantas-gds (approved):

infrastructure/template.yaml
huwd: ! Do'h I chaned that last min and forgot to update the other bits, great spot. Changing now

divyaparameswaran (approved):

PR #70, opened 2023-01-10
GUA-615: make prettifyDate formatter handle seconds and milliseconds

DougsCER:
What?
Make prettifyDate formatter handle seconds and milliseconds.
Despite only seeing dateEpoch values from TxMA in seconds we have coded defensively to handle milliseconds as well as that is a more standard format. 

Why?
During integration testing we found that the dateEpoch from TXMA event was in seconds not milliseconds. The prettifyDate dateformatter in format-user-services lambda was setup to only handle milliseconds so was always returning the wrong date.

### DIFF
--- a/lambda/format-user-services/format-user-services.ts
+++ b/lambda/format-user-services/format-user-services.ts
@@ -56,8 +56,16 @@ const validateTxmaEvent = (txmaEvent: TxmaEvent): void => {
 };
 
 export const prettifyDate = (dateEpoch: number): string => {
+  let dateEpochInMilliSeconds: number;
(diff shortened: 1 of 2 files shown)


divyaparameswaran (approved):

DougsCER:
Closing PR till HUW gets back and we decide what to do with the dates in the DB.  Not worth the manual testing of this work if it's never to be used

DougsCER:
Closing PR till HUW gets back and we decide what to do with the dates in the DB. Not worth the manual testing of this work if it's never to be used

PR #72, opened 2023-02-01
GUA-618: Add sonar coverage jobs to workflow.

divyaparameswaran:


huwd (approved):
Great work!

huwd:
I think GDS way might not have test coverage because we don't necessarily want to shot for a metric here.
The way just says [appropriate](https://gds-way.cloudapps.digital/manuals/code-review-guidelines.html#testing). I think that's based off a history in which shooting for a specific metric can lead to tests that are either a bit bizarre or of limited value, or some canny engineering to keep the 100% up, even if it doesn't always provider the confidence you might wish.

I tend to agree with posts like this:
https://dev.to/noriller/you-dont-need-100-coverage-412p

That if we can easily get to 100%, then sure why not.
But tests are also a trade off of time, and we should use that sparingly.

So we should ensure the most important tests are there, and not fall into a trap like expending 90% of our energies on the last 10% of tests, unless we're really sure of the value!

PR #89, opened 2023-04-25
BAU: Check for null values as well as absence of keys

huwd:
Extend validation to allow for that


lambda/save-raw-events/save-raw-events.ts
```diff
@@ -37,16 +37,19 @@ export const getTTLDate = (): number => {
 };
 
 export const validateUser = (user: UserData): void => {
-  if (user.user_id === undefined) {
+  if (user.user_id === undefined || user.user_id === null) {
```
TaherKapasi, line 40:
```suggestion
if (!user.user_id) {
'''

Comment (huwd) [reply lambda/save-raw-events/save-raw-events.ts]: Amazing, much better cheers for the improvement @TaherKapasi that's way less verbose!

**Review by TaherKapasi — APPROVED** [review]

**Review by DougsCER — APPROVED** [review]

File: lambda/save-raw-events/save-raw-events.ts [hunk 2]
```diff
@@ -37,17 +37,17 @@ export const getTTLDate = (): number => {
 };
 
 export const validateUser = (user: UserData): void => {
-  if (user.user_id === undefined) {
+  if (!user.user_id) {
```
divyaparameswaran, line 40: Nice!. We can do this on the other lambdas as well for consistency :)

divyaparameswaran (approved):

PR #90, opened 2023-04-25
GUA-746: Add Session ID to raw event store

huwd:
This is part of extending the raw events store to handle events we will need to render the activity log. We'll want to use the AUTH_AUTH_CODE_ISSUED event, but be able to aggregate them by the earliest event grouped by a session_id.


huwd:
Dependent on: https://github.com/alphagov/di-txma-event-self-serve/pull/105

divyaparameswaran (approved):

PR #95, opened 2023-05-17
GUA-748- Query Activty Log

huwd:
Adds a Lambda that:

Reads in an from the dummy dynamoDb stream which is a TxMA event, and  queries the activity_log table to get an entry for the user_id and a session_id . 
The response from the dynamo db could be empty in which case the lambda sends just the TXMA event.
If a response id received both the response and the TXMA event is returned back to the next lambda.
The output is written  to a subsequent SQS queue

Note: There is a concern regarding the security implications of the data stored on the raw events store. Hence we are using a dummy event store for now. Once the concerns are addressed we can update the query lambda to read from the raw events store stream.


lambda/query-activity-log/models.ts
```diff
@@ -0,0 +1,43 @@
+type UrnFdnSub = string;
```
DougsCER, line 1: why are some fields pulled out and others not?

lambda/query-activity-log/tests/query-activity-log.test.ts
```diff
@@ -0,0 +1,292 @@
...
+            },
+          },
+        ],
+      })
+    );
+    expect(() => {
+      validateTxmaEventBody(txmaEvent);
+    }).toThrowError();
+  });
+  test("throws error when timestamp is missing", () => {
```
DougsCER, line 106: all the test files i've seen we have space above each test and each describe.  I think it helps readability a lot

lambda/query-activity-log/tests/query-activity-log.test.ts
```diff
@@ -0,0 +1,292 @@
...
+        ],
+      })
+    );
+    expect(() => {
+      validateTxmaEventBody(txmaEvent);
+    }).toThrowError();
+  });
+  test("throws error when user_id  is missing", () => {
+    const txmaEvent = JSON.parse(
+      JSON.stringify({
```
DougsCER, line 160: This is ok but to be quicker, you can use object deconstruction and overrides to create tests parameters with altered fields 
Eg https://github.com/alphagov/di-account-management-backend/blob/main/lambda/save-raw-events/tests/save-raw-events.test.ts#L77


the syntax is 

`...<original object>, <field to mutate> : <new value>`


This means if the original object changes you don't need to re-write all the tests to update the object being created

lambda/query-activity-log/tests/query-activity-log.test.ts
```diff
@@ -0,0 +1,292 @@
...
+    }).toThrowError();
+  });
+});
+
+describe("sendSqsMessage", () => {
+  beforeEach(() => {
+    sqsMock.reset();
+    process.env.QUEUE_URL = queueUrl;
+  });
+  afterEach(() => {
```
DougsCER, line 209: spacing again, personally i like spacing between functions for readability

lambda/shared-models.ts
DougsCER, line 1: this file will be removed by other PR. 👍

lambda/query-activity-log/models.ts
divyaparameswaran: That's a good point not sure why it is needed . Happy for it to be all in the same format .

lambda/query-activity-log/tests/query-activity-log.test.ts
divyaparameswaran: I assumed my prettier plugin would take care of it . I will add the space .

lambda/query-activity-log/models.ts
```diff
@@ -0,0 +1,43 @@
...
+  event_type: string;
+  session_id: string;
+  user_id: string;
+  timestamp: number;
+  activities: Activity[];
+  truncated: boolean;
+}
+
+export interface UserActivityLog {
+  TxmaEvent: TxmaEvent;
```
DougsCER, line 36: type parameters should be lower case i think

divyaparameswaran: yup okay I will fix this here and we can add ticket for all the other lambdas as well .

lambda/query-activity-log/query-activity-log.ts
```diff
@@ -0,0 +1,126 @@
...
+  const { AWS_REGION } = process.env;
+  const client = new SQSClient({ region: AWS_REGION });
+  const message: SendMessageRequest = {
+    QueueUrl: queueUrl,
+    MessageBody: messageBody,
+  };
+  const result = await client.send(new SendMessageCommand(message));
+  return result.MessageId;
+};
+export const handler = async (event: DynamoDBStreamEvent): Promise<void> => {
```
DougsCER, line 93: space

DougsCER (approved):

PR #105, opened 2023-06-01
GUA 748 delete activity log

DougsCER:
Lambda to delete records in the activity log store, triggered by subscription to the account deletion SNS topic

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -253,9 +253,9 @@ Resources:
         SSEType: KMS
         KMSMasterKeyId: !GetAtt DatabaseKmsKey.Arn
 
-  ######################################
-  # User Activity Log  store
-  ######################################
+  #########################
+  # User Activity Log store
+  #########################
   UserActivityLogStore:
     Type: AWS::DynamoDB::Table
     DeletionPolicy: Delete
@@ -299,6 +299,7 @@ Resources:
         SSEEnabled: true
         SSEType: KMS
         KMSMasterKeyId: !GetAtt DatabaseKmsKey.Arn
+
   ######################################
   # TxMA Event Queue (Dummy for now)
   ######################################
@@ -308,6 +309,7 @@ Resources:
       KmsMasterKeyId: !GetAtt QueueKmsKey.Arn
     DeletionPolicy: Delete
     UpdateReplacePolicy: Delete
+
   ######################
   # Save Raw Events
   ######################
@@ -464,9 +466,9 @@ Resources:
         - !Ref AlarmNotificationTopic
       ActionsEnabled: true
 
-  ######################
-  # Querying the Record
-  ######################
+  ########################
+  # Querying User Services
+  ########################
   QueryUserServicesFunction:
     Type: AWS::Serverless::Function
     DependsOn:
@@ -642,9 +644,9 @@ Resources:
         - !Ref AlarmNotificationTopic
       ActionsEnabled: true
 
-  #######################
-  # Formatting the Record
-  #######################
+  #################################
+  # Formatting User Services Record
+  #################################
   FormatUserServicesFunction:
     Type: AWS::Serverless::Function
     DependsOn:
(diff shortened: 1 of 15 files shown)


lambda/delete-activity-log/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "eslint-plugin-import": "^2.26.0",
+    "eslint-plugin-prettier": "^4.2.1",
+    "jest": "^29.1.2",
+    "prettier": "^2.7.1",
+    "ts-jest": "^29.0.3",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.8.4"
+  },
+  "dependencies": {
+    "@aws-sdk/client-sqs": "^3.186.0",
```
divyaparameswaran, line 38: There is a vulnerability that is pulled as a part of this version of client-sqs https://github.com/alphagov/di-account-management-backend/pull/112/files

lambda/delete-activity-log/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@types/aws-lambda": "^8.10.106",
+    "@aws-sdk/client-dynamodb": "^3.186.0",
```
divyaparameswaran, line 40: Same as above

lambda/delete-activity-log/tests/delete-activity-log.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test("it iterates over each record in the batch", async () => {
+    await handler(TEST_SNS_EVENT);
+    expect(dynamoMock.commandCalls(DeleteCommand).length).toEqual(2);
+  });
```
divyaparameswaran, line 50: Could we add something similar to : https://github.com/alphagov/di-account-management-backend/pull/105/files#diff-fdfb65ad97796cd1e379b0543ddbc9f13e9940177fce4359209d250f5fef7232R29

divyaparameswaran (approved):

lambda/delete-activity-log/delete-activity-log.ts
```diff
@@ -0,0 +1,126 @@
...
+  userData: UserData
+): Promise<ActivityLogEntry[] | undefined> => {
+  const { TABLE_NAME } = process.env;
+  const command = {
+    TableName: TABLE_NAME,
+    KeyConditionExpression: "user_id = :user_id",
+    ExpressionAttributeValues: {
+      ":user_id": userData.user_id,
+    },
+    // TODO: make multiple requests when dynamo has truncated the response to 10MB
```
alex9smith, line 47: Could you make a ticket on the backlog so we don't forget about this?

infrastructure/template.yaml
```diff
@@ -1834,6 +1835,233 @@ Resources:
...
+          - Effect: Allow
+            Action:
+              - sqs:SendMessage
+            Resource: !GetAtt DeleteActivityLogDeadLetterQueue.Arn
+          - Effect: Allow
+            Action:
+              - dynamodb:DeleteItem
+              - dynamodb:Query
+              - dynamodb:GetItem
+              - dynamodb:BatchWriteItem
```
alex9smith, line 1974: My first reaction here was that this lambda shouldn't need any write permissions, but is this because there's no distinction between batching delete requests and batching write requests?

alex9smith (approved):
As before, just a couple of non-blocking questions for my understanding.

infrastructure/template.yaml
divyaparameswaran: Yes you are right Batch Write is used for DeleteRequests and PutRequests .

lambda/delete-activity-log/delete-activity-log.ts
divyaparameswaran: Yes will add thank you for spotting this .

PR #106, opened 2023-06-06
GUA 748 format activity log

divyaparameswaran:
Adds a Lambda that:

Reads in an from the sqs queue and formats the data into a structure to write in dynamodb
The output is written to a subsequent SQS queue

### DIFF
--- a/lambda/format-activity-log/format-activity-log.ts
+++ b/lambda/format-activity-log/format-activity-log.ts
@@ -1,0 +1,122 @@
+import { SQSEvent } from "aws-lambda";
+
+import {
+  SendMessageCommand,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import {
+  Activity,
+  ActivityLogEntry,
+  TxmaEvent,
+  UserActivityLog,
+  UserData,
+} from "./models";
+
+const createNewActivityLogEntryFromTxmaEvent = (
+  txmaEvent: TxmaEvent
+): ActivityLogEntry => ({
+  event_type: txmaEvent.event_name,
+  session_id: txmaEvent.user.session_id,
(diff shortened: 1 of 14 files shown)


lambda/format-activity-log/format-activity-log.ts
DougsCER, line 1: Would this be the correct place to truncate the log if there were too many activities in it?

DougsCER: the alternative is we don't truncate in the DB, but we do in the frontend

DougsCER: agreed on slack this lambda should not allow more than 100 activities in each record.  Upon receipt of activity 101  "truncated" field should be set to "true".
When further events are received no DB update should be performed.

divyaparameswaran: Thank you I have added this and test .

lambda/format-activity-log/format-activity-log.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+};
+
+export const formatIntoActivitLogEntry = (
+  userActivityLog: UserActivityLog
+): ActivityLogEntry => {
+  if (userActivityLog.activityLogEntry === undefined) {
+    return createNewActivityLogEntryFromTxmaEvent(userActivityLog.txmaEvent);
+  }
+  if (userActivityLog.activityLogEntry.activities.length < 99) {
```
DougsCER, line 78: why less than 99?  should it not be < 100

DougsCER: to get a maximum of 100 activities in an event

divyaparameswaran: `Oh yes . Not sure what I was thinking

PR #117, opened 2023-06-28
GUA-887: Update query to continue till all records are read

divyaparameswaran:
- Dynamo Db paginates the query at 10 MB , this process will make sure that if there are more records all of them are queried before a deletion is performed .


alex9smith (approved):
I agree, this is a perfectly sensible time to ignore the 'no await in a loop' rule 👍🏻

lambda/delete-activity-log/delete-activity-log.ts
DougsCER, line 1: is it possible to add a unit test for this?

PR #125, opened 2023-07-06
OLH-902 : Update Logging to exclude logging personal data from txma events

divyaparameswaran:
Update lambdas to not log any data from the TXMA event.

### DIFF
--- a/lambda/delete-activity-log/delete-activity-log.ts
+++ b/lambda/delete-activity-log/delete-activity-log.ts
@@ -29,9 +29,7 @@ export const validateUserData = (userData: UserData): UserData => {
   if (userData.user_id) {
     return userData;
   }
-  throw new Error(
-    `userData did not have a user_id: ${JSON.stringify(userData)}`
-  );
+  throw new Error(`userData did not have a user_id`);
 };
 
 export const getAllActivitiesoForUser = async (
--- a/lambda/delete-activity-log/delete-activity-log.ts
+++ b/lambda/delete-activity-log/delete-activity-log.ts
@@ -126,13 +126,15 @@ export const handler = async (event: SNSEvent): Promise<void> => {
           await batchDeleteActivityLog(activityRecords);
         }
       } catch (err) {
-        console.error(err);
         const message: SendMessageRequest = {
           QueueUrl: DLQ_URL,
           MessageBody: record.Sns.Message,
         };
-        console.log("sending message to SQS %s", JSON.stringify(message));
-        await sqsClient.send(new SendMessageCommand(message));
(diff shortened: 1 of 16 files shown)


lambda/query-activity-log/query-activity-log.ts
```diff
@@ -60,9 +60,7 @@ export const validateTxmaEventBody = (txmaEvent: TxmaEvent): void => {
   ) {
     validateUser(txmaEvent.user);
   } else {
-    throw new Error(
-      `Could not validate UserServices ${JSON.stringify(txmaEvent)}`
```
DougsCER, line 64: non PII data from the TXMA event could still be helpful to triage issues. eg the client_id, event_name.
Could this be logged?

divyaparameswaran: So these messages go to the DLQ and we log the message ID there. We can get the entire paylod if we look at the message in the DLQ ? WOuld it be helpful to still log client_id and event_name

divyaparameswaran: Ah sorry we don't log message Id for messages sent to DLQ maybe that would be a better option ?

DougsCER: yeah, message ID seems key.  If the error log gives us message ID and we can see the message content from DLQ then that's grand

lambda/delete-activity-log/delete-activity-log.ts
```diff
@@ -133,8 +131,10 @@ export const handler = async (event: SNSEvent): Promise<void> => {
           QueueUrl: DLQ_URL,
           MessageBody: record.Sns.Message,
         };
-        console.log("sending message to SQS %s", JSON.stringify(message));
-        await sqsClient.send(new SendMessageCommand(message));
+        const result = await sqsClient.send(new SendMessageCommand(message));
+        console.log(
+          `[Message sent to DLQ] with message id = ${result.MessageId}`
```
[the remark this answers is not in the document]
divyaparameswaran: Yes it still logs Validation Error i.e Could not validate User. Then we can check the message to see the payload to fifure out what the missing piece was ?

DougsCER (approved):

lambda/delete-activity-log/delete-activity-log.ts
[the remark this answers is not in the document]
DougsCER: Chatted to Divya on call. 
She pointed out this is being logged by a log line above that's not shown in this PR summary. 

I've approved on the basis that the log lines are moved to be next to eachother

PR #127, opened 2023-08-03
OLH-1006: Save the event ID in the activity log

alex9smith:
I was checking that this was possible and I thought the easiest way to make sure was to do the work.

There's three commits here, changing the query, format and write lambdas in turn. It ended up being a fairly simple change but see the commit messages for more information.

### DIFF
--- a/lambda/format-activity-log/format-activity-log.ts
+++ b/lambda/format-activity-log/format-activity-log.ts
@@ -13,6 +13,13 @@ import {
   UserData,
 } from "./models";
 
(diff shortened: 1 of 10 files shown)


divyaparameswaran:
Based on the figma this morning we report a single Sign in event which could have multiple activities . Would that mean we send a list of all the event_ids in a Sign in event to the ticketing system ?

DougsCER (approved):

alex9smith:
@divyaparameswaran yes I think so. If a user is reporting a session as suspicious we'll want to send the session ID and all the event IDs in that session to counter fraud for investigation

PR #131, opened 2023-08-08
BAU: Update Sonar to the latest action version

alex9smith:
This commit SHA is v2.0.0 of the action - see https://github.com/SonarSource/sonarcloud-github-action/releases/tag/v2.0.0 We've been getting warnings that the version we were on before was using Java 11 which is very out of date (Java 17 is the current release).

Updating should get us a much newer version

### DIFF
--- a/.github/workflows/sonarcloud.yaml
+++ b/.github/workflows/sonarcloud.yaml
@@ -27,7 +27,7 @@ jobs:
         run: yarn test:cov
 
       - name: SonarCloud Scan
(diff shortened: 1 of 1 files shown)


alex9smith:
Phew, no warnings now on the Sonar comments (see [here](https://github.com/alphagov/di-account-management-backend/pull/128#issuecomment-1664094873) for an example of a comment with the warning).

DougsCER (approved):

PR #143, opened 2023-09-20
OLH-904: Decrypt user activities before formatting

alex9smith:
### What changed

Update the `query-activity-log` lambda to decrypt the user's activity history in the query result from DynamoDB. 

This closely follows the implementation of the encryption added in #142 with a few simplifications:

1. We don't need the backup key to decrypt. This meant I could simplify the keyring builder
2. This doesn't use the check value as part of the encryption context validation. I will add this later

### Why did it change

We need to decrypt the user's history before we try to insert the new event. It made sense to do it in this Lambda since it's already making the query to DynamoDB - this lets us keep the format Lambda 'permissionless'. 

### Related links

#142 

- [x] Added new environment variable

### Permissions

## How to review

I am working on a test checklist to verify this in dev

### DIFF
--- a/lambda/query-activity-log/decrypt-data.ts
+++ b/lambda/query-activity-log/decrypt-data.ts
@@ -1,0 +1,23 @@
+import { KmsKeyringNode, buildDecrypt } from "@aws-crypto/client-node";
+import buildKmsKeyring from "./kms-keyring-builder";
+
+const MAX_ENCRYPTED_DATA_KEY = 5;
+const DECODING = "utf8";
+
+const decryptClientConfig = { maxEncryptedDataKeys: MAX_ENCRYPTED_DATA_KEY };
+const decryptClient = buildDecrypt(decryptClientConfig);
+
+let keyring: KmsKeyringNode;
+
+async function decryptData(data: string): Promise<string> {
+  try {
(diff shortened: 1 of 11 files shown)


lambda/query-activity-log/decrypt-data.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const MAX_ENCRYPTED_DATA_KEY = 5;
+const DECODING = "utf8";
+
+const decryptClientConfig = { maxEncryptedDataKeys: MAX_ENCRYPTED_DATA_KEY };
+const decryptClient = buildDecrypt(decryptClientConfig);
+
+let keyring: KmsKeyringNode;
+
+export function generateExpectedContext(userId: string): EncryptionContext {
```
DougsCER, line 16: so the accessCheckValue is not being checked?
VC store use this as an additional confirmation.  I'm not sure how much more security it really provides. 
If we're not going to use it to validate then we shouldn't write it

DougsCER (approved):

lambda/query-activity-log/decrypt-data.ts
alex9smith: I'm going to add that check in a future PR, but I wanted to keep this one as the MVP for decryption

PR #155, opened 2023-11-09
OLH-1213 Add Dynatrace layer to backend lambdas

andrew-moores:
## Proposed changes

Add Dynatrace to backend resources.

### What changed

Updated template as per Observability teams documentation.
Moved common values to Globals and switched to using the public VPC subnets.

### Why did it change

To allow metrics to be shipped to Dynatrace.

### Permissions

## Testing

Data can be seen in Dynatrace for the dev environment.

## How to review

When deployed in dev data from interactions with the Account Home page can be seen in Dynatrace.

The process to add Dynatrace is documented in this README https://github.com/govuk-one-login/observability-infrastructure/tree/main/fargate.
You also need to know about this https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3809542559/Dynatrace+VPC+Endpoint page which is not mentioned in the README.

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -98,6 +98,37 @@ Conditions:
 
 Globals:
   Function:
+    Environment:
+      Variables:
+        AWS_LAMBDA_EXEC_WRAPPER: /opt/dynatrace
+        DT_CONNECTION_AUTH_TOKEN: !Sub
+          - '{{resolve:secretsmanager:${SecretArn}:SecretString:DT_CONNECTION_AUTH_TOKEN}}'
+          - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn ]
+        DT_CONNECTION_BASE_URL: !Sub
+          - '{{resolve:secretsmanager:${SecretArn}:SecretString:DT_CONNECTION_BASE_URL}}'
+          - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn ]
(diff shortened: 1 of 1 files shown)


infrastructure/template.yaml
```diff
@@ -98,6 +98,37 @@ Conditions:
...
+    Architectures: ["arm64"]
+    MemorySize: 128
+    KmsKeyArn: !GetAtt LambdaKMSKey.Arn
+    ReservedConcurrentExecutions: 10
+    Runtime: nodejs18.x
+    Timeout: 5
+    VpcConfig:
+      SubnetIds:
+          - Fn::ImportValue: !Sub "${VpcStackName}-ProtectedSubnetIdA"
+          - Fn::ImportValue: !Sub "${VpcStackName}-ProtectedSubnetIdB"
```
domantas-gds, line 129: Just to confirm are deployments only set across two available zones? I know there's an option for three. You should be able to confirm the number of AZs available based on the CloudFormation VPC AvailabilityZoneCount parameter. Not necessarily in scope of Dynatrace work but just worth noting.

andrew-moores: Yes, its set to 2 in dev, build, integration and staging.  I've raised https://govukverify.atlassian.net/browse/OLH-1281 to consider using all three AZs/

alex9smith: That's right. The first version of the VPC stack only exported two subnets which is what we originally set up on. At somepoint between then and now they've added the option for the third AZ

alex9smith (approved):

PR #159, opened 2023-11-22
OLH-1242: Set up SNS topic for reporting suspicious activity

aenciso:
## Proposed changes

Set up a new SNS topic for the ‘report suspicious activity' messages in the backend template and store the topic ARN in a new secrets manager parameter.

### What changed

Added section in the template to create both resources

### DIFF
--- a/infrastructure/template.yaml
+++ b/infrastructure/template.yaml
@@ -1058,6 +1058,13 @@ Resources:
       KmsMasterKeyId: !Ref SnsKmsKey
       TopicName: UserAccountDeletion
 
+  ReportSuspiciousActivityTopic:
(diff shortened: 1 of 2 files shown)


alex9smith (changes requested):
Could you switch this to a Systems Manager Parameter instead of using Secrets Manager please? 

Secrets Manager costs money per secret but parameters are free (up to a large limit) and this topic ARN doesn't need to be protected as a secret. 

You can look at the definition for our other SNS topic for how to define a parameter: 
https://github.com/alphagov/di-account-management-backend/blob/main/infrastructure/template.yaml#L1100-L1106

alex9smith (approved):
Nice one, thanks!

PR #161, opened 2023-11-28
OLH-1267: Add ADR 0010 - Simplify activity log data structure & pipeline

alex9smith:
### What changed

Add ADR 0010 - Simplify activity log data structure & pipeline. ([Link to rendered document](https://github.com/alphagov/di-account-management-backend/blob/OLH-1267-change-activity-log-structure/docs/adr/0010-simplify-activity-log-data-structure-pipeline.md))

This ADR proposes a simplified data structure and pipeline architecture that will allow this table to be more flexible with changing requirements. 

### Why did it change

We've been asked to consider some new requirements for the activity log - eg. allowing users to report individual or multiple events at once. Without making a change to the data structure it will be difficult to meet these new requirements at any point, but if we make them now we can continue to work with the current requirements with the option to add new features in the future. 

### Related links

This partially supersedes ADRs 0006 and 0009


Viji-GDS-2023 (approved):
All looks good to me

PR #162, opened 2023-11-28
OLH-1049 Send an audit event to TxMA when a user submits a report

peterfajemisincabinetoffice:
## Proposed changes
OLH-1049 Send an audit event to TxMA when a user submits a report
 We create a Lambda that’s subscribed to the suspicious activity SNS topic, so that when the lambda receives a message, it will generate the audit event and send it to the audit event SQS queue. TxMA will pick up messages from the Audit Event Queue

### What changed

1. A new lambda that sends events to TxMA - This handler for this Lambda is built to be generic such that when another Lambda needs to send a message to TxMA(i.e. put a message on the audit event SQS queue, tall they will need to do is add a validation rule, add their transformation logic to convert the received event to a TxMA event and configure their Lambda within the infrastructure code, the transformation logic and validation rule is selected by an event name, so as part of the Lambda, the new event name would need to be configured as an environment variable.
2. A generic Validator function that will validate an event based on rules.
3. Error handling to ensure failures messages are sent to the associated DLQ

### Why did it change
So that reported suspicious activities can sent to TxMA

### Related links
https://govukverify.atlassian.net/browse/OLH-1049
https://govukverify.atlassian.net/browse/OLH-857
https://govukverify.atlassian.net/browse/OLH-1242
https://github.com/alphagov/di-account-management-backend/blob/main/docs/adr/0008-report-a-suspicious-activity.md

## Checklists
- The lambda should be able to subscribe to the SNS topic.
- The lambda should validate the fields .
- If validation fails send the message to the DLQ.
- If validation is a success , the lambda should then send the event to the audit event queue

- EVENT_NAME - This must always be configured against any new lambda that uses the handler.
- [ ] No secrets were added or changed

## Testing
- Deployed to Dev
- Publish a suspicious activity message to the SNS topic
- Verify a TxMA valid event was sent to the  audit event SQS queue
- Verify a failure e.g. due to validation error as sent to the SendSuspiciousActivityToTxMADeadLetterQueue

### DIFF
--- a/lambda/send-event-to-txma/send-event-to-txma.ts
+++ b/lambda/send-event-to-txma/send-event-to-txma.ts
@@ -1,0 +1,124 @@
+import { SNSEvent } from "aws-lambda";
+import {
+  SendMessageCommand,
+  SendMessageCommandOutput,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import { TxmaEvent } from "./models";
+import {
+  COMPONENT_ID,
+  EventNamesEnum,
+  REPORT_SUSPICIOUS_ACTIVITY_EVENT_NAME,
+  ValidationRulesKeyEnum,
+} from "./constants";
+import VALIDATOR_RULES_MAP from "./validator-rules";
+import validateObject from "./validator";
+
+export const transformToTxMAEvent = (event: any, eventName: string): any => {
+  let txmaEvent = null;
+  if (eventName === EventNamesEnum.HOME_REPORT_SUSPICIOUS_ACTIVITY) {
+    txmaEvent = {
+      component_id: COMPONENT_ID,
+      event_name: REPORT_SUSPICIOUS_ACTIVITY_EVENT_NAME,
+      user: {
+        user_id: event.user_id,
+        persistent_session_id: event.persistent_session_id,
+        session_id: event.session_id,
+      },
+      extensions: {
+        reported_session_id: event.reported_event.session_id,
+      },
+    };
+  } else {
+    throw new Error(
+      "Unsupported event - There is no transformation logic for this event"
+    );
+  }
+  return txmaEvent;
+};
+
+export const sendSqsMessage = async (
+  messageBody: string,
+  queueUrl: string | undefined
+): Promise<SendMessageCommandOutput> => {
+  const { AWS_REGION } = process.env;
+  const client = new SQSClient({ region: AWS_REGION });
+  const message: SendMessageRequest = {
+    QueueUrl: queueUrl,
+    MessageBody: messageBody,
+  };
+  return client.send(new SendMessageCommand(message));
+};
+/**
+ * A function for sending an audit event message to the TxMA SQS queue.
+ *
+ * @param txmaEvent - the transformed txma event
+ * @param queueUrl
+ * @returns the response returned by the SQS client
+ * @throws error
+ * Thrown when an exception is thrown while trying to send a message to the queue
+ */
+export async function sendAuditEvent(
+  txmaEvent: TxmaEvent,
+  queueUrl: string | undefined
+): Promise<SendMessageCommandOutput> {
+  try {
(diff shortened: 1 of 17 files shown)


lambda/send-event-to-txma/send-event-to-txma.ts
```diff
@@ -0,0 +1,122 @@
...
+    throw error;
+  }
+}
+
+export const handler = async (event: SNSEvent): Promise<void> => {
+  const { DLQ_URL, EVENT_NAME, TXMA_QUEUE_URL } = process.env;
+  await Promise.all(
+    event.Records.map(async (record) => {
+      try {
+        if (EVENT_NAME) {
```
andrew-moores, line 81: Consider reversing the logic of this statement and turning it into a guard condition check.

lambda/send-event-to-txma/send-event-to-txma.ts
```diff
@@ -0,0 +1,122 @@
...
+          } else {
+            throw new Error(
+              `Received Event: ${JSON.stringify(
+                receivedEvent
+              )} failed validation.`
+            );
+          }
+        } else {
+          throw new Error(
+            "Event name must be provided as an environment variable"
```
andrew-moores, line 110: Consider stating how this error has impacted the outcome of this function:

"Cannot handle event as event name has not been provided in the environment"

lambda/send-event-to-txma/send-event-to-txma.ts
```diff
@@ -0,0 +1,122 @@
...
+export const handler = async (event: SNSEvent): Promise<void> => {
+  const { DLQ_URL, EVENT_NAME, TXMA_QUEUE_URL } = process.env;
+  await Promise.all(
+    event.Records.map(async (record) => {
+      try {
+        if (EVENT_NAME) {
+          const receivedEvent: any = JSON.parse(record.Sns.Message);
+          if (
+            validateObject(receivedEvent, VALIDATOR_RULES_MAP.get(EVENT_NAME))
+          ) {
```
andrew-moores, line 85: Similar to last comment, reversing the logic will make this a guard condition.

lambda/send-event-to-txma/send-event-to-txma.ts
```diff
@@ -0,0 +1,122 @@
...
+            const txmaEvent = transformToTxMAEvent(receivedEvent, EVENT_NAME);
+            if (
+              validateObject(
+                txmaEvent,
+                VALIDATOR_RULES_MAP.get(ValidationRulesKeyEnum.TXMA_EVENT)
+              )
+            ) {
```
andrew-moores, line 92: Also consider making this a guard condition.

lambda/send-event-to-txma/validator-rules.ts
andrew-moores, line 1: Could we have done this with JSON schema?

andrew-moores (commented):
Is there a requirement to report how many of these events we emit?  We aren't logging anything to show when an event is published so would have to look somewhere else if this was required.

lambda/send-event-to-txma/send-event-to-txma.ts
peterfajemisincabinetoffice: Agreed, now a guard condition

lambda/send-event-to-txma/send-event-to-txma.ts
peterfajemisincabinetoffice: Agreed, updated

peterfajemisincabinetoffice:
> Is there a requirement to report how many of these events we emit? We aren't logging anything to show when an event is published so would have to look somewhere else if this was required.

I have added similar logging to other Lambdas that post to queue, there is always a generic log showing message being posted to Queue. No requirement for count as per ticket

lambda/send-event-to-txma/validator-rules.ts
peterfajemisincabinetoffice: Yes we could have. However this would require getting security to authorise. Discussed with Alex and longer term view of contract validation would means we can eventually pull out the in lambda validations. So happy to stay with current validation mechanism.

PR #173, opened 2024-01-03
OLH-1296 remove the query-activity-log lambda and DLQ.

andrew-moores:
## Proposed changes

Removing the query-activity-log lambda.

### What changed

The SAM template was updated to remove the lambda, its DLQ, roles, etc.
GitHub actions updated to no longer reference the lambda.

### Why did it change

No longer needed.

### Permissions

## Testing

Deployment to dev.


docs/adr/0010-simplify-activity-log-data-structure-pipeline.md
```diff
@@ -22,7 +22,6 @@ The new structure for each item in DynamoDB will be:
 | `reported_suspicious` |               | boolean   | If the user has reported this event as suspicious. This will always be false on initial write. |
 
 This data structure is much simpler than before which allows us to simplify the pipeline.
-We no longer need to query DynamoDB for an existing item before updating it, so we can remove the `query-activity-log` Lambda.
```
alex9smith, line 25: Can we leave this line in please? The ADR should be a snapshot of the decision we made at the time we made it so we've got a history of why we did this.

alex9smith (approved):
Preemptively approving this because I won't need to take another look once you've addressed my comment

PR #174, opened 2024-01-05
OLH-1298 Modify the write-activity-log lambda to use the new data model.

andrew-moores:
## Proposed changes

Use the new data model to write to the activity log table.

### What changed

wite-activity-log accepts new data model as input from format-activity-log lambda and writes new format to DynamoDB table.  The event_id is envelope encrypted.

### Why did it change

We have a new data mode for the activity log.

### Permissions

### DIFF
--- a/lambda/write-activity-log/write-activity-log.ts
+++ b/lambda/write-activity-log/write-activity-log.ts
@@ -10,25 +10,12 @@ import {
   SendMessageRequest,
   SQSClient,
 } from "@aws-sdk/client-sqs";
-import {
-  Activity,
-  ActivityLogEntry,
-  EncryptedActivityLogEntry,
-} from "./models";
+import { ActivityLogEntry, EncryptedActivityLogEntry } from "./models";
 import encryptData from "./encrypt-data";
 
 const dynamoClient = new DynamoDBClient({});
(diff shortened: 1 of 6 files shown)


lambda/write-activity-log/write-activity-log.ts
```diff
@@ -73,17 +61,17 @@ export const handler = async (event: SQSEvent): Promise<void> => {
...
         validateActivityLogEntry(activityLogEntry);
-        const encryptedActivities: string = await encryptData(
-          JSON.stringify(activityLogEntry.activities),
-          activityLogEntry.user_id
-        );
         const encryptedActivityLog: EncryptedActivityLogEntry = {
-          event_type: activityLogEntry.event_type,
+          event_id: activityLogEntry.event_id,
+          event_type: await encryptData(
+            activityLogEntry.event_type,
```
andrew-moores, line 67: @alex9smith is it correct that we only want to encrypt the event type?

alex9smith: Yes that's right. Viji and I discussed with security and we agreed that was an appropriate level of defense vs. performance.

alex9smith (approved):

PR #175, opened 2024-01-08
OLH-1297 - Modify the format-activity-log Lambda to use the new data model

peterfajemisincabinetoffice:
## Proposed changes

Use the new data model, such that when new activity log TXMA Events are added to the dynamo, we format correctly as per new data model before adding to the Queue for write to the activity log table.

JIRA: https://govukverify.atlassian.net/browse/OLH-1297

### What changed
Picks up TXMA event of activity log as per new model and generate an activity log that is consumable by the write activity log lambda.

### Why did it change
To align with the new data model for thew activity log.

## Testing

1. Navigate to dev environment
2. Add a new valid record to the Dynamo DB with TxMA Event structure as per new data model
3. Should add an ActivityLogEntry to the ActivityLogToWriteQueue
4. Add a new invalid record to  Dynamo DB
5. Should add an entry to the FormatActivityLogDeadLetterQueue

Sample correct new TxMA Event as per new model
`{"event_id":"12345678","timestamp":123456789,"timestamp_ms_formatted":"x","timestamp_ms":1234,"event_name":"AUTH_AUTH_CODE_ISSUED","client_id":"client-id-value","session_id":"123456789","user_id":"qwerty","reported_suspicious":true}
  `

Sample correct output ActivityLogEntry as per new model is:
`{
  "event_id": "12345678",
  "event_type": "AUTH_AUTH_CODE_ISSUED",
  "session_id": "123456789",
  "user_id": "qwerty",
  "client_id": "client-id-value",
  "timestamp": 123456789,
  "reported_suspicious": true
}`

## How to review
Code completeness  and quality, tests

### DIFF
--- a/lambda/format-activity-log/format-activity-log.ts
+++ b/lambda/format-activity-log/format-activity-log.ts
@@ -1,4 +1,6 @@
-import { SQSEvent } from "aws-lambda";
+import { DynamoDBStreamEvent } from "aws-lambda";
+import { AttributeValue } from "@aws-sdk/client-dynamodb";
+import { unmarshall } from "@aws-sdk/util-dynamodb";
 
 import {
   SendMessageCommand,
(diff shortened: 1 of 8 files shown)


lambda/format-activity-log/models.ts
```diff
@@ -21,15 +23,22 @@ export interface Activity {
 }
```
andrew-moores, line 23: Can activity be deleted as its not used in ActivityLogEntry now.

lambda/format-activity-log/models.ts
```diff
@@ -21,15 +23,22 @@ export interface Activity {
...
   timestamp: number;
-  activities: Activity[];
-  truncated: boolean;
+  client_id: string;
+  reported_suspicious: boolean;
 }
 
 export interface UserActivityLog {
   txmaEvent: TxmaEvent;
   activityLogEntry: ActivityLogEntry | undefined;
```
andrew-moores, line 37: Can activityLogEntry ever be undefined now?

lambda/format-activity-log/format-activity-log.ts
```diff
@@ -1,79 +1,45 @@
...
   event_type: txmaEvent.event_name,
-  session_id: txmaEvent.user.session_id,
-  user_id: txmaEvent.user.user_id,
+  session_id: txmaEvent.session_id,
+  user_id: txmaEvent.user_id,
+  client_id: txmaEvent.client_id,
   timestamp: txmaEvent.timestamp,
-  activities: [activityFromTxmaEvent(txmaEvent)],
-  truncated: false,
+  reported_suspicious: false,
```
andrew-moores, line 21: Should this always be false?  Could we introduce a constant with a name that clarifies why this is always false.

lambda/format-activity-log/format-activity-log.ts
```diff
@@ -90,23 +56,33 @@ export const sendSqsMessage = async (
...
+          );
+          console.log(`[Message sent to QUEUE] with message id = ${messageId}`);
+        } else {
+          console.log(
+            `DB stream sent a ${txmaEvent.event_name} event. Irrelevant for activity log so ignoring`
+          );
+        }
       } catch (err) {
-        const messageId = await sendSqsMessage(record.body, DLQ_URL);
+        const messageId = await sendSqsMessage(JSON.stringify(record), DLQ_URL);
```
andrew-moores, line 85: What happens if this call fails?  Do we lose the error logging below and the details of the event that we couldn't handle?

peterfajemisincabinetoffice: Logging that a failure happens before attempting to send to DLQ, with error, but not including the associated event ID, as this would reqquire unmarshalling which can also fail. Also not including the entire DB stream record as potentially sensitive data.

lambda/format-activity-log/format-activity-log.ts
peterfajemisincabinetoffice: Yes default false at this stage, addressed by introduced a default const.

lambda/format-activity-log/models.ts
peterfajemisincabinetoffice: Addressed , in this intance, UserActivityLog interface is not even needed at all, removed.

andrew-moores (approved):

PR #185, opened 2024-01-30
Olh 1111 restructure backend lambdas

andrew-moores:
## Proposed changes

Restructure the backend lambdas.

### What changed

The structure has been changed so that each lambda handler lives in the /src folder with shared /src/modules in common and tests in /tests.
Moved all the config files and the SAM template and config into the root folder.
Updated GitHub actions to reflect new structure and new way to package lambda handlers for deployment.
Add single sonarcloud configuration.

### Why did it change

To make the lambda handlers part of the same node project to simplify the development and allow easy code sharing.

### Permissions

## Testing

All the unit tests are running correctly.
The SAM template was deployed to dev.
The GitHub actions have been run after pushes to the branch and when requesting the deployment to dev.
Coverage report and static analysis can now be seen in Sonarcloud.

## How to review

Please check the lambdas work as expected and that the unit tests still test what was intended as the code base has changed a lot.


.github/workflows/sonarcloud.yaml
```diff
@@ -7,13 +7,10 @@ on:
   pull_request:
```
domantas-gds, line 7: Can narrow to relevant types of pull requests

```
  pull_request:
    types: [ opened, synchronize, reopened ]
```

.github/workflows/sonarcloud.yaml
```diff
@@ -7,13 +7,10 @@ on:
...
-  sonarcloud_query_user_services:
-    name: query-user-services
+  sonarcloud_home_backend:
+    name: home_backend
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    defaults:
-      run:
-        working-directory: ./lambda/query-user-services
     steps:
```
domantas-gds, line 14: Some of these steps may not be required, around npm dependencies and coverage reporting. Docs here provide a code snippet with required steps https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3668280149/How+to+integrate+GitHub+repositories+with+SonarCloud

Which I've also done here https://github.com/govuk-one-login/account-management-stubs/blob/main/.github/workflows/sonarcloud-oidc-token.yaml

May be an improvement after initial refactor but one to note

saralk:
I was thinking a bit about the file structure, currently you have it like this...

```
- src
   - delete-activity-log.ts
   - delete-email-subscription.ts
```

I think something like this could be a bit better long term...

```
- src
   - delete-activity-log
      - handler.ts
   - delete-email-subscription.ts
      - handler.ts
```

With a directory for each lambda, you can separate the code into multiple files for more complex lambdas.

.github/workflows/sonarcloud.yaml
andrew-moores: Done.

.github/workflows/sonarcloud.yaml
andrew-moores: I have tidied up the github workflow for sonarqube.  If I remove the step that generates the coverage report, as per your referenced config, then the code coverage is flagged as 0%,

andrew-moores:
@saralk That does sound like a good idea for when we do have more complex lambdas.  I'd hope we don't need to build anything that merits more than the handler module and some modules in common.  If we do get to the point where we want to do this then we can introduce this structure then.

peterfajemisincabinetoffice:
> @andrew-moores ran locally a few times, something seems to be recreating the infrastructure directory and placing the build.toml file in it.

Ignore this, it was files hangong around from previous builds

peterfajemisincabinetoffice:
LGTM, although seems to be attempting to run a some additional checks that are no longer required.

saralk (approved):
Code wise it looks good, as long as people's testing doesn't throw up any issues, it looks good to go!

peterfajemisincabinetoffice (approved):

andrew-moores:
I removed the branch protection rules for the three specific lambdas.

aenciso (approved):

PR #186, opened 2024-01-30
OLH-1053 update reported flag on activity

saralk:
## Proposed changes

When a user reports an activity as suspicious, we should record that in the activity log, so that we can ensure we don't report things twice, and provide feedback to the user.

### What changed

- Create a new lambda, and all the infrastructure that goes along with that (log groups, policies etc...), in the CloudFormation template
- Add a secondary index to the `activity_log` table, so that we can query the table based on the `event_id`
- When the new lambda is triggered, we retrieve the event from the event log (based on the event_id), and update it, setting `reported_suspicious` to `true`

### Why did it change

When we build functionality for users to report activity as suspicious, we want to a) give feedback to the user, and b) ensure that we don't make multiple tickets in zendesk.

By setting this flag to true, we can show that to the user, so that they know we have received the report. This will work even if there is an issue with Zendesk, and we haven't got a ticket number yet.

### Related links

https://github.com/govuk-one-login/di-account-management-backend/pull/181

- [X] Added parameters to Cloudformation template
- [X] Added new environment variable

## Testing

I manually tested this by...

Deploying the changes to the dev environment

Getting an `event_id` from the `activity_log` table (through the AWS console)
Screenshot 2024-01-30 at 14 22 01

Publishing a message on the `SuspiciousActivity` topic with the following body (replacing event_id with the one I got in the previous step)
```
{
  "event_id": "b62c6ec9-30e6-4f7d-9769-02e25b81baf6",
  "event_name": "event_name",
  "timestamp": 1609462861,
  "timestamp_formatted": "timestamp_formatted",
  "client_id": "client_id",
  "user": {
    "user_id": "user_id",
    "session_id": "session_id",
    "govuk_signin_journey_id": "govuk_signin_journey_id"
  },
  "reported_suspicious": true
}
```

Refreshing the DyanamoDB view, and checking that the `reported_suspicious` property has been set to true for the event I chose.

### DIFF
--- a/src/mark-suspicious-activity-as-reported.ts
+++ b/src/mark-suspicious-activity-as-reported.ts
@@ -9,6 +9,7 @@ import {
   DynamoDBDocumentClient,
   UpdateCommand,
   QueryCommand,
+  QueryCommandOutput,
 } from "@aws-sdk/lib-dynamodb";
 import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
 import { ActivityLogEntry } from "./common/model";
@@ -43,11 +44,11 @@ export const getItemByEventId = async (
     },
   });
 
-  const result = await dynamoDocClient.send(getItem);
+  const result: QueryCommandOutput = await dynamoDocClient.send(getItem);
 
-  if (result?.Items?.length !== 1) {
+  if (result.Items?.length !== 1) {
     throw Error(
-      `Expecting exactly 1 result from getItemByEventId, but got ${result?.Items?.length}`
+      `Expecting exactly 1 result from getItemByEventId, but got ${result.Items?.length}`
     );
   }
   const item = result.Items[0];
@@ -87,6 +88,11 @@ export const handler = async (event: SNSEvent): Promise<void> => {
         if (!INDEX_NAME) {
           throw new Error(
             "Cannot handle event as index name has not been provided in the environment"
+          );
+        }
+        if (!DLQ_URL) {
+          throw new Error(
+            "Cannot handle event as DLQ url has not been provided in the environment"
           );
         }
         const receivedEvent: ActivityLogEntry = JSON.parse(record.Sns.Message);
--- a/src/mark-suspicious-activity-as-reported.ts
+++ b/src/mark-suspicious-activity-as-reported.ts
@@ -1,0 +1,109 @@
+import { SNSEvent } from "aws-lambda";
+import {
+  SendMessageCommand,
+  SendMessageCommandOutput,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import {
+  DynamoDBDocumentClient,
+  UpdateCommand,
+  QueryCommand,
+} from "@aws-sdk/lib-dynamodb";
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
+import { ActivityLogEntry } from "./common/model";
+
+const dynamoClient = new DynamoDBClient({});
+const dynamoDocClient = DynamoDBDocumentClient.from(dynamoClient);
+
+export const sendSqsMessage = async (
+  messageBody: string,
+  queueUrl: string | undefined
+): Promise<SendMessageCommandOutput> => {
+  const { AWS_REGION } = process.env;
+  const client = new SQSClient({ region: AWS_REGION });
+  const message: SendMessageRequest = {
+    QueueUrl: queueUrl,
+    MessageBody: messageBody,
+  };
+  return client.send(new SendMessageCommand(message));
+};
+
+export const getItemByEventId = async (
+  tableName: string,
+  indexName: string,
+  eventId: string
+): Promise<{ user_id: string; timestamp: number }> => {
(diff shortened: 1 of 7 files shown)


alex9smith (commented):
This looks really good - nice one! 

I'd forgotten this table still had a composite primary key of `user_id` and `timestamp` which I think has made your work more complicated than it should have been. Sorry about that. We should have already updated the table first to use the new primary key of `user_id` and `event_id` we defined in [ADR 0010](https://github.com/govuk-one-login/di-account-management-backend/blob/main/docs/adr/0010-simplify-activity-log-data-structure-pipeline.md). You've done the right thing here in working with the primary key as it it but it does need to change for this to be robust as we should expect to see multiple events for a user with the same timestamp at some point since timestamp is only to the nearest second.

I think what we should do is update the indices now in your PR. Then you could get rid of the `getItemByEventId` function and update `markEventAsReported` to use the `user_id` and `event_id` keys. We can rely on DynamoDB to enforce uniqueness for the primary key. We'll also be able to get rid of the global secondary indices which will save us a bit of money.

The first part of the table definition would then look like this: 

```yaml
 UserActivityLogStore:
    Type: AWS::DynamoDB::Table
    DeletionPolicy: Delete
    UpdateReplacePolicy: Delete
    Properties:
      AttributeDefinitions:
        - AttributeName: user_id
          AttributeType: S
        - AttributeName: event_id
          AttributeType: S
      BillingMode: PAY_PER_REQUEST
      TableName: !Ref ActivityLogStoreTableName
      KeySchema:
        - AttributeName: user_id
          KeyType: HASH
        - AttributeName: event_id
          KeyType: RANGE
```

 Again, sorry for making this harder than it needed to be! Let me know if you want to chat about any of this in more detail.

saralk:
Make sense, thanks @alex9smith.

I wonder if it's worth doing that in a separate ticket, as there are a places in other lambdas where the timestamp is used in a query ([here](https://github.com/govuk-one-login/di-account-management-backend/blob/olh-1053-update-reported-flag-on-activity/lambda/delete-activity-log/delete-activity-log.ts#L79-L86) for example). So the scope of this PR would increase.

alex9smith:
Ah yeah - that's a good point. I'd forgotten this was still in the delete lambda as well so I thought it'd still be only changes to this lambda. I'll write up a ticket to change the primary key for next sprint.

src/mark-suspicious-activity-as-reported.ts
```diff
@@ -0,0 +1,109 @@
...
+        const receivedEvent: ActivityLogEntry = JSON.parse(record.Sns.Message);
+
+        const { user_id, timestamp } = await getItemByEventId(
+          TABLE_NAME,
+          INDEX_NAME,
+          receivedEvent.event_id
+        );
+        await markEventAsReported(TABLE_NAME, user_id, timestamp);
+      } catch (err) {
+        const response = await sendSqsMessage(record.Sns.Message, DLQ_URL);
```
andrew-moores, line 101: What happens if we can't send to the DLQ?  Environment variable might not be set, might have the wrong name or service is down.

src/mark-suspicious-activity-as-reported.ts
```diff
@@ -0,0 +1,109 @@
...
+    ExpressionAttributeValues: {
+      ":event_id": eventId,
+    },
+  });
+
+  const result = await dynamoDocClient.send(getItem);
+
+  if (result?.Items?.length !== 1) {
+    throw Error(
+      `Expecting exactly 1 result from getItemByEventId, but got ${result?.Items?.length}`
```
andrew-moores, line 50: We know we have a length defined at this point.
```suggestion
      `Expecting exactly 1 result from getItemByEventId, but got ${result.Items.length}`
```

src/mark-suspicious-activity-as-reported.ts
```diff
@@ -0,0 +1,109 @@
...
+          TABLE_NAME,
+          INDEX_NAME,
+          receivedEvent.event_id
+        );
+        await markEventAsReported(TABLE_NAME, user_id, timestamp);
+      } catch (err) {
+        const response = await sendSqsMessage(record.Sns.Message, DLQ_URL);
+        console.error(
+          `[Message sent to DLQ] with message id = ${response.MessageId}`,
+          err as Error
```
andrew-moores, line 104: If we can't send to the DLQ we will also lose the original error that caused us to try to send to the DLQ.  Suggest err is logged first in the catch block.

saralk: If there is an error that stops us reaching this line of code, then it will throw an error which will be caught in the logs

src/mark-suspicious-activity-as-reported.ts
saralk: `Items` is an optional field, so potentially can be `undefined`, so the optional chaining here stops a typescript error

andrew-moores: You're right, Items can still be undefined all we know is that it isn't 1.

andrew-moores (approved):

PR #191, opened 2024-02-07
OLH-1389 Redesign the backend for reporting a suspicious activity

andrew-moores:
## Proposed changes

New ADR documenting decision to use Step Functions.

### What changed

New ADR added that documents the decision to use Step Functions.

### Why did it change

The current approach is not as good as that offered by Step Functions.

### Permissions

### DIFF
--- a/docs/adr/0011-reporting-suspicious-activity.md
+++ b/docs/adr/0011-reporting-suspicious-activity.md
@@ -1,0 +1,98 @@
+# Design for reporting suspicious activity
+
+Establish a pattern for implementing workflows in a resilient and extendable way.  The initial use case will be the report suspicious activity workflow.
+
+## Context
+
+We will allow the user to select one or more records from the activity log to report as suspicious activity.  A suspicious activity report consists of a Zendesk ticket, a TxMA event and an email confirmation.  All of these tasks must finish before a suspicious activity report is considered complete.  Some of the tasks must happen serially and some can be run in parallel.  All of the tasks must be resilient to error. 
+
+The required tasks are:
+
+1. Update the activity log items to show report requested.
+2. Create Zendesk ticket
+3. Update the activity log items with the ticket number
+4. Send an event to TxMA that includes the Zendesk ticket number
+5. Update the activity log items with the TxMA event id
+6. Send an email to Notify.
+7. Update the activity log items with id of request to Nofify.
+
+Notes.
+
+* Step 1 allows the front end to redisplay the activity log with the status of the reported activity records shown correctly to prevent multiple reports of the same activity.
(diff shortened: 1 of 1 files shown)


andrew-moores:
@alex9smith and @andrew-moores to review.

docs/adr/0011-reporting-suspicious-activity.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Option 3 - single lambda
+
+The whole process could be handled by a single lambda that picks up an event from SQS to decouple the front end.  The lambda would then perform each step required in turn:
+
+1. Update the database.
+2. Create the ticket.
+3. Store the ticket id in the database.
+4. Publish a TxMa event.
+5. Send an email via Notify.
```
saralk, line 88: I'm not sure that this order is correct. I would have thought tickets 1, 4 and 5 could all be done in parallel (and have no dependencies on other tickets being done). Steps 2 and 3 are the only ones that have a dependency.

andrew-moores: I made updating the database the first task as we need to ensure a revisit to the activity page provides the correct information about the report.  Agree we could run tasks 1 and 2 in parallel.

docs/adr/0011-reporting-suspicious-activity.md
```diff
@@ -0,0 +1,98 @@
...
+We will allow the user to select one or more records from the activity log to report as suspicious activity.  A suspicious activity report consists of a Zendesk ticket, a TxMA event and an email confirmation.  All of these tasks must finish before a suspicious activity report is considered complete.  Some of the tasks must happen serially and some can be run in parallel.  All of the tasks must be resilient to error. 
+
+The required tasks are:
+
+1. Update the activity log items to show report requested.
+2. Create Zendesk ticket
+3. Update the activity log items with the ticket number
+4. Send an event to TxMA that includes the Zendesk ticket number
+5. Update the activity log items with the TxMA event id
+6. Send an email to Notify.
```
peterfajemisincabinetoffice, line 16: As per yesterday's discussion, do we still need the notify steps?

andrew-moores: No we don't but I think having this documented as part of the process still makes a good case for adopting Step Functions.  And the requirements around the emailing are still changing.

alex9smith: Latest update from Ben and TICF is that we're going to switch back to using Notify now, but only send the confirmation email. Let's leave this step in the ADR for now and if we don't end up building it we can consider coming back later and removing it if it's confusing.

alex9smith (approved):

PR #194, opened 2024-02-20
OLH-1437 Send confirmation e-mail to users when they report suspicious activity

saralk:
## Proposed changes

This PR adds a new lambda to the backend which will send an e-mail to the user, confirming receipt of their suspiciouus activity report.

### What changed

* Added a new lambda SendConfEmail which sends an email (using GovUK Notify)
* Subscribe the lambda to the SuspiciousActivity topic
* Copied the client registry from the [translations file](https://github.com/govuk-one-login/di-account-management-frontend/blob/main/src/locales/en/translation.json) in the frontend

### Why did it change

We needed to copy the translation file from the frontend, as the e-mails require us to show the human readable name of the service that is being reported.

We discussed various options, and this seems like the most tactical in terms of getting something released within the timeframe. Long terms we would look to do something more robust, like having the registry stored in DynamoDB (and have the frontend read from it).

I decided to the copy the entire client registry object from the frontend over, even though it contains fields we wont use in the email. My thinking is that it is less error prone to copy and paste the entire object over from the frontend.

There is still a bit of [follow up work](https://govukverify.atlassian.net/browse/OLH-1500) to do, which is to make this a step in the [step function work ](https://govukverify.atlassian.net/browse/OLH-1454) that Peter is doing. Once we do that, we'll also know what the Zendesk ticket ID is, so we can put that in the email.

- [X] Added parameters to Cloudformation template
- [X] Added new secret or systems manager parameter
- [X] Added new environment variable

### Permissions

## Testing

I have deployed this to the development environment, and published an event on the `SuspiciousActivity` SNS topic with the following payload...

```
{"user":{"email":"saral.kaushik@digital.cabinet-office.gov.uk"},"client_id":"lite","timestamp":1708348905253}
```

I received the following email 

Screenshot 2024-02-21 at 09 14 53


src/@types/notifications-node-client.d.ts
```diff
@@ -0,0 +1,67 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
saralk, line 1: got from here https://github.com/alphagov/notifications-node-client/issues/175

src/send-conf-email.ts
```diff
@@ -0,0 +1,127 @@
...
+        assert(notifyApiKey, `${NOTIFY_API_KEY} secret not retrieved`);
+
+        const receivedEvent: SuspiciousActivityEvent = JSON.parse(
+          record.Sns.Message
+        );
+
+        await sendConfMail(notifyApiKey as string, TEMPLATE_ID, receivedEvent);
+      } catch (err) {
+        console.error(`Error sending email for event`, err);
+        const response = await sendSqsMessage(record.Sns.Message, DLQ_URL);
```
andrew-moores, line 119: Do we need to catch exceptions sending to the DLQ?

peterfajemisincabinetoffice: This point was raised before and the rationale at the time is that we want to be able to log an error if send to DLQ fails

peterfajemisincabinetoffice: Although not sure if this is a pattern we are following across all our lambdas at present, so approving on the basis of I am not passionate about it.

peterfajemisincabinetoffice (approved):
LGTM

src/send-conf-email.ts
andrew-moores: If the sendSqsMessage fails by throwing an exception then it will escape here and we won't log anything.

PR #206, opened 2024-03-12
OLH-1566 Send time and date as one string to Notify

saralk:
## Proposed changes

Send the fully formatted date and time string to Notify, instead of a seperate date string and time string,

### Why did it change

[Context from Dana](https://gds.slack.com/archives/C012GEZBZM0/p1710157972567999?thread_ts=1710155049.156179&cid=C012GEZBZM0)

I was wondering whether there is any advantage of passing event date and event time into the Notify template separately vs passing through the entire date formatted with [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) (which we use[ on the frontend](https://github.com/govuk-one-login/di-account-management-frontend/blob/172beee57f197eb33947b324919f49d8d87666ae/src/utils/prettifyDate.ts#L22), not sure whether you can do the same on the backend but I would have imagined yes since it’s still Javascript?) which should automatically output our desired format for English and Welsh when the correct arguments are used. So then the activity details in our notify template could look something like this, and we wouldn’t need to hard code the “at” in Welsh:  

> ((eventName))
> ((eventDate)) (UK time)

### DIFF
--- a/src/send-conf-email.ts
+++ b/src/send-conf-email.ts
@@ -32,16 +32,11 @@ export const getClientInfo = (
 
 const formatTimestamp = (timestamp: number, language: string) => {
   const date = new Date(timestamp * 1000);
-  return {
-    date: Intl.DateTimeFormat(language, {
-      dateStyle: "long",
-      timeZone: "Europe/London",
-    }).format(date),
-    time: Intl.DateTimeFormat(language, {
-      timeStyle: "short",
-      timeZone: "Europe/London",
-    }).format(date),
-  };
+  return Intl.DateTimeFormat(language, {
+    timeStyle: "short",
+    dateStyle: "long",
+    timeZone: "Europe/London",
+  }).format(date);
 };
 
 export const formatActivityObjectForEmail = (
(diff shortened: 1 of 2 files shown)


danacotoran:
Hi @saralk, thanks for considering my suggestion regarding the format of date/time for Notify.

This is not really a review on the changes on this PR but just a comment/observation/nitpick on the name of the files. Wondering if the name should be more specific and referencing reporting suspicious activity in some way. It's not really an issue, it's just that it might need renaming in a hypothetical future where we use lambdas to send other types of confirmation emails anyway. 
But maybe we will never use lambdas to send any other confirmation emails.

saralk:
@danacotoran that's a fair point, I suppose if we need to send other confirmation emails, then it would be easy enough to rename it then

danacotoran:
Cool, wasn't sure how easy it'd be to rename at a later time once it's all working in prod. Sounds like it's an easy thing so in that case don't mind me 👍🏼

alex9smith (approved):

PR #211, opened 2024-03-14
OLH-1592 - use date of reported activity in confirmation email.

andrew-moores:
## Proposed changes
When informing the user about the activity they reported we should use the date of the activity rather than the date they reported it.

### What changed
Use the date of the activity being reported to populate the state of the StepFunction and use that to populate the confirmation email.

### Why did it change
The date being used was the date the activity was reporting which would have confusing to the reporter.

### Permissions


saralk (approved):

peterfajemisincabinetoffice (approved):
LGTM

PR #212, opened 2024-03-14
OLH-1593: Include hint text in emails for One Login Home

alex9smith:
### What changed

Include hint text in emails for One Login Home. 

This does hard code the client IDs for the Home frontend, but we have all the rest of the content duplicated here anyway so I didn't think that was too bad. 

I'm using Notify's [Optional content feature](https://www.notifications.service.gov.uk/using-notify/optional-content) to show / hide the hint text line for events about One Login Home.

This relies on a template update which I've made in the Test Notify service. Once the PR is merged I'll apply the same template update to the live service

### Why did it change

To explain to users what 'Your GOV.UK One Login Home' means

## Testing

I've deployed this to dev and triggered an email. I've also checked that the hint text doesn't show up when the variable is set to false.

image

image


andrew-moores (approved):
This looks good.  I did notice in the template you have `((showHomeHintText??^))` for when the hint text shouldn't be displayed.  Is this needed?  Couldn't check the prod template as I don't have privs to view.

alex9smith:
Ah no, that's a separate optional block that adds a newline in the inset text block if the hint text is present to make the spacing better

PR #214, opened 2024-03-18
OLH-1594 Refactor error handling in send-conf-email

aenciso:

### Permissions


src/send-conf-email.ts
```diff
@@ -128,6 +128,39 @@ export const handler = async (
     return input;
   } catch (err) {
     console.error(`Error sending email for event`, err);
```
alex9smith, line 130:
```suggestion
```
Looking good! Could you also remove this line so we're never logging the raw error message unless we're at the last branch in `notifyErrorHandler`?

alex9smith (approved):

PR #228, opened 2024-06-04
OLH-1816-add-a-code-deploy-service-role-to-our-backend-stack

latifcabinetoffice:
## Proposed changes

Enabling canary deployment for our backend lambda's to observe behaviour.

### What changed

1. Formatted automatically caused by prettier on commit.
2. Conditional to enable canary deployment on dev only.
3. Addition of deployment alarms for all back end lambda's to check for errors.
4. Addition of Code deploy role with Lambda permissions.
5. Enabled Lambda alias auto published with "live".
6. Configured canary deployment to deploy 10% of instances for 5 mins, if there are no errors deploy to all.

### Why did it change

To automate rollback in case of errors in newly deployed lambda.

### Related links

https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3823436177/Lambda+-+Canary+Deployment+Migration+Guidance#Pre-requisites

### Permissions

## Testing

1. Validated lambda displays live alias.
2. On deployment validated 10% of lambda's have the new version of code.
3. Validated new alarms are created.
4. Validated a roll back caused by manually triggering alarm.

## How to review

1. This will add 5 mins to dev deployment time to see your new code. Opinions on a different frequency e.g. 25% every 1 minute will cut full changes to 4 mins and you are more likely to hit a instance with your new code earlier. (for a separate ticket)
2. Suggestions on custom alarm metrics as we are currently just looking at Lambda Errors which may not be the best. (for a separate ticket)

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -181,7 +181,7 @@ Globals:
         Fn::If:
           - IsDevelopment
           - Canary10Percent5Minutes
-          - !Ref LambdaCanary10Percent5Minutes
+          - AllAtOnce
       Role: !GetAtt CodeDeployServiceRole.Arn
 
 Mappings:
@@ -2941,17 +2941,6 @@ Resources:
           !Ref AWS::NoValue,
         ]
 
-  LambdaCanary10Percent5Minutes:
-    Type: AWS::CodeDeploy::DeploymentConfig
-    Properties:
-      ComputePlatform: Lambda
-      DeploymentConfigName: LambdaCanary10Percent5Minutes
-      TrafficRoutingConfig:
-        TimeBasedCanary:
-          CanaryPercentage: 10
-          CanaryInterval: 5
-        Type: TimeBasedCanary
-
   CreateSupportTicketsDeploymentAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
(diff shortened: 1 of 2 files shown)


template.yaml
```diff
@@ -84,6 +84,11 @@ Conditions:
           - !Ref Environment
           - "dev"
 
+  IsDevelopment:
```
peterfajemisincabinetoffice, line 87: Please use the existing IsnotDevelopment and use NOT, some examples here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html

template.yaml
```diff
@@ -137,8 +167,22 @@ Globals:
...
+        - "{{resolve:secretsmanager:${SecretArn}:SecretString:NODEJS_LAYER}}"
+        - SecretArn:
+            !FindInMap [
+              EnvironmentConfiguration,
+              !Ref Environment,
+              dynatraceSecretArn,
+            ]
+    AutoPublishAlias: live
+    DeploymentPreference:
+      Enabled: true
```
peterfajemisincabinetoffice, line 179: please change the else action here to the all at once option

template.yaml
```diff
@@ -2799,6 +2916,495 @@ Resources:
...
+        !If [
+          UsePermissionsBoundary,
+          !Ref PermissionsBoundary,
+          !Ref AWS::NoValue,
+        ]
+
+  LambdaCanary10Percent5Minutes:
+    Type: AWS::CodeDeploy::DeploymentConfig
+    Properties:
+      ComputePlatform: Lambda
```
peterfajemisincabinetoffice, line 2947: Do you still need this?

latifcabinetoffice: removed

template.yaml
latifcabinetoffice: not possible to change

peterfajemisincabinetoffice (approved):

PR #261, opened 2024-07-02
BAU: Correct typo in condition definition

alex9smith:
### What changed

We changed these definitions in #260 but Cloudformation is picky and `Equals` must be capitalised.

It also doesn't always like the short form `!Or` function when we have a definition spanning multiple lines, so change that to the full `Fn::Or` form. 

I'm very surprised the linter action didn't catch this - I'll investigate that separately.

### Why did it change

Currently our deploys are failing in build. 

- [x] ~Added~ Edited parameters in Cloudformation template

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -81,21 +81,22 @@ Conditions:
 
   IsNotDevelopmentOrDemo:
     Fn::Not:
-      - Fn::Or: 
-        - Fn::equals:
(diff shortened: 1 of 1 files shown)


aenciso (approved):

shannonwhitworth (approved):
nice :)

PR #269, opened 2024-07-03
[OLH-1940] Enable skippable canary builds

saralk:
### What changed

Add the ability to disable canary builds by adding `[skip canary]` to the commit message

We do this by following the instructions [here](https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3836051600/Rollback+Recovery+Guidance#Escape-Hatch%3A-how-to-skip-canary-deployments-when-needed)

### Why did it change

There may be a need to disable canary builds sometimes as they slow down deployment, for example during an incident.

### Related links

- [Guide for enabling this](https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3836051600/Rollback+Recovery+Guidance#Escape-Hatch%3A-how-to-skip-canary-deployments-when-needed)
- https://govukverify.atlassian.net/browse/OLH-1817

### Permissions

## Testing

I added a clause to deploy this branch on push, and verified that it deploys using the AllAtOnce directive

Screenshot 2024-07-10 at 12 08 37

I also verified that the build succeeded and the lambdas don't timeout by manually triggering the Report Suspicious Activity step function

### DIFF
--- a/src/save-raw-events.ts
+++ b/src/save-raw-events.ts
@@ -88,7 +88,7 @@ export const handler = async (event: SQSEvent): Promise<void> => {
         };
         const result = await sqsClient.send(new SendMessageCommand(message));
         console.error(
-          `[Message sent to DLQ] aaa with message id = ${result.MessageId}`,
+          `[Message sent to DLQ] with message id = ${result.MessageId}`,
           err,
         );
       }
--- a/src/save-raw-events.ts
+++ b/src/save-raw-events.ts
@@ -21,7 +21,7 @@ const translateConfig = { marshallOptions };
 const dynamoClient = new DynamoDBClient({});
 const dynamoDocClient = DynamoDBDocumentClient.from(
   dynamoClient,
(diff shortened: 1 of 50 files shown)


src/save-raw-events.ts
```diff
@@ -88,7 +88,7 @@ export const handler = async (event: SQSEvent): Promise<void> => {
         };
         const result = await sqsClient.send(new SendMessageCommand(message));
         console.error(
-          `[Message sent to DLQ] with message id = ${result.MessageId}`,
+          `[Message sent to DLQ] aaa with message id = ${result.MessageId}`,
```
latifcabinetoffice, line 91: Possibly accidentally added?

latifcabinetoffice (changes requested):

src/save-raw-events.ts
saralk: yes it was, good spot!

PR #278, opened 2024-07-08
OLH-1949: Set up a basic post-deploy integration test

alex9smith:
### What changed

Set up to run a basic post-deploy integration test. This is following the dev platform team's [instructions](https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3054010402/How+to+run+tests+against+your+deployed+application+in+a+SAM+deployment+pipeline).

The test script uses the AWS cli to invoke the `write-activity-log` and`delete-activity-log` lambdas with pre-defined fixtures. It'll fail with a non-zero exit code if either invocation fails.

This is a super-simple integration test, but it checks a lot:
- The lambdas have built correctly and can run
- They have permission to use KMS and DynamoDB
- They can process the message without timing out

Notably this script won't fail if any of the payloads are invalid due to the way we catch and handle errors in the lambdas. That's ok for now.

In the future we'll probably want to migrate this to an NPM script using something like Cucumber to orchestrate the tests. This is a first go at getting a post-deploy test working; it'll catch the most broken bad releases as a start.

The default role the post-deploy tests run as doesn't have many permissions. The pipeline provides the `TestRoleArn` parameter to the deployed stack so we can build our own policy and grant it the
permissions we need.

After this is merged and the first testing image has been pushed to ECR I will update the pipeline settings to run the tests in build.

### Why did it change

We've had a couple of incidents where we've broken the lambda builds and release non-functional artifacts. We'll want to expand on this test suite to cover more of the functions of the backend, but this is a first effort to catch the type of releases that are have caused the recent problems. 

### Permissions

## Testing

I've built the Docker container and run the test script against build using my AWS credentials. You can do the same by running:

```bash
aws sso login --profile di-account-build-admin

eval "$(aws configure export-credentials --profile di-account-build-admin --format env)"

docker build . -t test

docker run -t \
  -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
  -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
  -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN \
  -e AWS_SECURITY_TOKEN=$AWS_SECURITY_TOKEN \
  -e AWS_DEFAULT_REGION="eu-west-2" \
  test
```

The real test will be with the role's credentials once this is deployed. There's no way to test that now however.

### DIFF
--- a/post-deploy-tests/fixtures/write-activity-log.json
+++ b/post-deploy-tests/fixtures/write-activity-log.json
@@ -13,8 +13,8 @@
       "messageAttributes": {},
       "md5OfBody": "e359050680237e70f4b71093185eca98",
       "eventSource": "aws:sqs",
-      "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:MyQueue",
-      "awsRegion": "us-east-1"
+      "eventSourceARN": "arn:aws:sqs:eu-west-2:123456789012:MyQueue",
+      "awsRegion": "eu-west-2"
     }
   ]
 }
--- a/post-deploy-tests/fixtures/write-activity-log.json
+++ b/post-deploy-tests/fixtures/write-activity-log.json
@@ -1,0 +1,20 @@
+{
+  "Records": [
+    {
+      "messageId": "19dd0b57-b21e-4ac1-bd88-01bbb068cb78",
+      "receiptHandle": "MessageReceiptHandle",
+      "body": "{\"event_id\":\"event-id-1\",\"event_type\":\"AUTH_AUTH_CODE_ISSUED\",\"session_id\":\"session-id\",\"user_id\":\"post-deploy-tests\",\"timestamp\":1720465374,\"client_id\":\"client-id\",\"reported_suspicious\":false}",
+      "attributes": {
+        "ApproximateReceiveCount": "1",
+        "SentTimestamp": "1523232000000",
+        "SenderId": "123456789012",
+        "ApproximateFirstReceiveTimestamp": "1523232000001"
+      },
+      "messageAttributes": {},
+      "md5OfBody": "e359050680237e70f4b71093185eca98",
+      "eventSource": "aws:sqs",
+      "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:MyQueue",
+      "awsRegion": "us-east-1"
+    }
+  ]
+}
--- a/.github/workflows/sam-app-post-merge-actions.yaml
+++ b/.github/workflows/sam-app-post-merge-actions.yaml
@@ -55,7 +55,7 @@ jobs:
 
       - name: Login to Amazon ECR
         id: login-ecr
-        uses: aws-actions/amazon-ecr-login@v1
+        uses: aws-actions/amazon-ecr-login@062b18b96a7aff071d4dc91bc00c4c1a7945b076 #pin@v2.0.1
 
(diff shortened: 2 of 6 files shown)


post-deploy-tests/fixtures/write-activity-log.json
```diff
@@ -0,0 +1,20 @@
...
+        "ApproximateReceiveCount": "1",
+        "SentTimestamp": "1523232000000",
+        "SenderId": "123456789012",
+        "ApproximateFirstReceiveTimestamp": "1523232000001"
+      },
+      "messageAttributes": {},
+      "md5OfBody": "e359050680237e70f4b71093185eca98",
+      "eventSource": "aws:sqs",
+      "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:MyQueue",
+      "awsRegion": "us-east-1"
```
aenciso, line 17: Would it matter if we use this instead of eu-west-2?

alex9smith: It doesn't seem to matter - I think this is indicating that the event originated in `us-east-1` but the lambda is still invoked in `eu-west-2` where it's deployed.

andyloughran: Let's try and keep eu-west-2 everywhere.  It doesn't matter, but it's just easier cognitively to know the vast majority is eu-west-2.

post-deploy-tests/fixtures/write-activity-log.json
```diff
@@ -0,0 +1,20 @@
...
+        "ApproximateReceiveCount": "1",
+        "SentTimestamp": "1523232000000",
+        "SenderId": "123456789012",
+        "ApproximateFirstReceiveTimestamp": "1523232000001"
+      },
+      "messageAttributes": {},
+      "md5OfBody": "e359050680237e70f4b71093185eca98",
+      "eventSource": "aws:sqs",
+      "eventSourceARN": "arn:aws:sqs:eu--west-2:123456789012:MyQueue",
+      "awsRegion": "eu--west-2"
```
aenciso, line 17:
```suggestion
      "eventSourceARN": "arn:aws:sqs:eu-west-2:123456789012:MyQueue",
      "awsRegion": "eu-west-2"
```

post-deploy-tests/fixtures/write-activity-log.json
```diff
@@ -0,0 +1,20 @@
...
+      "attributes": {
+        "ApproximateReceiveCount": "1",
+        "SentTimestamp": "1523232000000",
+        "SenderId": "123456789012",
+        "ApproximateFirstReceiveTimestamp": "1523232000001"
+      },
+      "messageAttributes": {},
+      "md5OfBody": "e359050680237e70f4b71093185eca98",
+      "eventSource": "aws:sqs",
+      "eventSourceARN": "arn:aws:sqs:eu--west-2:123456789012:MyQueue",
```
aenciso, line 16:
```suggestion
      "eventSourceARN": "arn:aws:sqs:eu-west-2:123456789012:MyQueue",
```

post-deploy-tests/fixtures/write-activity-log.json
alex9smith: Oh oops, thanks

aenciso (approved):

PR #310, opened 2024-07-23
OLH-1989-bau-refactor-backend-lambdas

latifcabinetoffice:
## Proposed changes
[OLH-1989] - Refactor backend lambda's

### What changed
Few minor refactors to:
- make code more readable
- removed duplicate code
- error for DLQ
- env var checks

### Why did it change
quick changes observed during another ticket

## Testing
unit tests pass

[OLH-1989]: https://govukverify.atlassian.net/browse/OLH-1989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/delete-activity-log.ts
+++ b/src/delete-activity-log.ts
@@ -5,13 +5,8 @@ import {
   WriteRequest,
 } from "@aws-sdk/client-dynamodb";
 import { DynamoDBDocumentClient, QueryCommand } from "@aws-sdk/lib-dynamodb";
-import {
-  SendMessageCommand,
-  SendMessageRequest,
-  SQSClient,
-} from "@aws-sdk/client-sqs";
 import { ActivityLogEntry, UserData } from "./common/model";
-import assert from "node:assert";
+import { sendSqsMessage } from "./common/sqs";
 
 const marshallOptions = {
   convertClassInstanceToMap: true,
@@ -23,8 +18,6 @@ const dynamoDocClient = DynamoDBDocumentClient.from(
   dynamoClient,
   translateConfig
 );
-
-const sqsClient = new SQSClient({});
 
 export const validateUserData = (userData: UserData): UserData => {
   if (userData.user_id) {
@@ -116,16 +109,19 @@ export const batchDeleteActivityLog = async (
 };
 
 export const handler = async (event: SNSEvent): Promise<void> => {
-  const { DLQ_URL, TABLE_NAME } = process.env;
-
   await Promise.all(
     event.Records.map(async (record) => {
+      const { DLQ_URL, TABLE_NAME } = process.env;
+      if (!DLQ_URL) {
+        throw new Error("DLQ_URL environment variable is not set");
+      }
+      if (!TABLE_NAME) {
+        throw new Error("TABLE_NAME environment variable is not set");
+      }
       try {
         console.log(
           `started processing message with ID: ${record.Sns.MessageId}`
         );
-        assert(DLQ_URL);
-        assert(TABLE_NAME);
 
         const userData: UserData = JSON.parse(record.Sns.Message);
         validateUserData(userData);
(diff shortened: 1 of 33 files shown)


alex9smith:
Looks like Github has got a bit confused with the change of base here. I'll sort that - it's a bit tricky to review like it is.

src/delete-activity-log.ts
```diff
@@ -116,16 +109,19 @@ export const batchDeleteActivityLog = async (
 };
 
 export const handler = async (event: SNSEvent): Promise<void> => {
-  const { DLQ_URL, TABLE_NAME } = process.env;
-
   await Promise.all(
     event.Records.map(async (record) => {
+      const { DLQ_URL, TABLE_NAME } = process.env;
+      if (!DLQ_URL) {
```
alex9smith: We have this pattern a lot now. Do you think it'd be worth adding a new helper to get an environment variable? Something like this
```typescript
export const getEnvironmentVariable(name: string): string => {
  const value = process.env[name];
  if (!value) {
    throw new Error(`${name} environment variable is not set`);
  }
};
```

src/delete-activity-log.ts
```diff
@@ -137,17 +133,16 @@ export const handler = async (event: SNSEvent): Promise<void> => {
...
-          QueueUrl: DLQ_URL,
-          MessageBody: record.Sns.Message,
-        };
-        const result = await sqsClient.send(new SendMessageCommand(message));
-        console.error(
-          `[Message sent to DLQ] with message id = ${result.MessageId}`,
-          err
-        );
+      } catch (error) {
+        console.error(`[Error occurred]: ${(error as Error).message}`);
```
alex9smith, line 136: Remind me - is there any risk `error.message` could contain part of the message payload? Our messages may contain personal data which we can't store in Cloudwatch.

src/tests/write-user-services.test.ts
```diff
@@ -138,23 +140,21 @@ describe("validateUserServices", () => {
 
 describe("validateServices", () => {
```
alex9smith, line 141:
```suggestion
describe("validateService", () => {
```

src/format-activity-log.ts
```diff
@@ -75,7 +61,7 @@ export const handler = async (event: DynamoDBStreamEvent): Promise<void> => {
         if (allowedTxmaEvents.includes(txmaEvent.event_name)) {
           validateTxmaEventBody(txmaEvent);
           const formattedRecord = formatIntoActivityLogEntry(txmaEvent);
-          const messageId = await sendSqsMessage(
+          const { MessageId: messageId } = await sendSqsMessage(
```
alex9smith, line 64: Nice!

src/delete-activity-log.ts
alex9smith: I see in a later commit you moved the validation of these values into the try-catch block. I think it's probably correct to not do that - if we're missing a critical environment variable then we can fail early and like you said on Slack we don't need to catch every error. The Lambda service can pick up on the error state and handle the retries and DLQ for us.

alex9smith (commented):
This is looking like a really good cleanup. I've left some suggestions mainly about how we get and validate environment variables. Let me know what you think!

src/delete-activity-log.ts
latifcabinetoffice: I believe it could be as we didn't log anything before, but removing would also make it difficult to figure out why our lambda failed... any suggestions? @alex9smith

alex9smith: We need to be 100% confident we're not logging personal data, so I think in this case we have to accept that our logs may not be as useful. An error message like "Could not process message ${messageID}" might be the best we can do - that'll at least allow us to trace the logs back to a message and replay it.

We can always add debug logging to a lower environment to diagnose an issue if we need to.

alex9smith (approved):
🚀

PR #312, opened 2024-07-25
OLH-1948-add-alarms-to-our-backend-for-lambda-success-rate-and-a-backlog-of-messages-on-sqs

latifcabinetoffice:
## Proposed changes
[OLH-1948] Add Lambda success rate alarm and alarm for oldest message in a queue

### What changed
Lambda:
- Check invocations vs errors to trigger an alarm when more than 5% trigger the alarm

Queue:
- If oldest item in queue is older than 5 minutes trigger the alarm 

### Why did it change
Improve alerting of errors

## Testing
Deployed to Dev:
- Observed alarm for a queue with messages longer than 5 minutes
- Graphing for lambda's with error rate > 5%

## How to review
Currently have an alarm for all lambda's and queues, double check if all are required i.e. Audit Output Queue. This will prevent notification fatigue. 

[OLH-1948]: https://govukverify.atlassian.net/browse/OLH-1948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -393,6 +393,7 @@ Resources:
   ######################################
   # Dummy events store
   ######################################
+
   DummyEventStore:
     Type: AWS::DynamoDB::Table
     DeletionPolicy: Delete
@@ -433,6 +434,7 @@ Resources:
   ######################################
   # Raw events store
   ######################################
+
   RawEventsStore:
     Type: AWS::DynamoDB::Table
     DeletionPolicy: Delete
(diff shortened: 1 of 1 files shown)


alex9smith (approved):
This looks good. I think we should keep the alarms on the audit output queue too. I'm not 100% sure what TxMA's SLA is but it's definitely under 1 minute so our 5 minute threshold makes sense.

PR #323, opened 2024-08-01
OLH-2025-throw-errors-in-backend-lambdas

latifcabinetoffice:
## Proposed changes

[OLH-2025]: Reconfigure the backend lambda's so that they all throw error's in the event of failures

### What changed

Within the handler of each lambda there is a try catch that explicitly throws an error.
Removed manual SQS queue push. 
NOTE: We recognise that this could cause batch processes to be duplicated.

### Why did it change

Missing errors from our lamba's as they didn't throw errors. 
This change works well with canary deployments which will allow for automatic roll backs by tracking the success rate alarms.

## Testing

Error raised in test
image

Proof of two retries:
image

Proof message was received by DLQ:
image

## How to review

[OLH-2025]: https://govukverify.atlassian.net/browse/OLH-2025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/create-support-ticket.ts
+++ b/src/create-support-ticket.ts
@@ -189,13 +189,8 @@ export const handler = async (
     console.log(`finished processing event with ID: ${eventIdentifier}`);
     return input;
   } catch (error) {
-    console.error(
-      `[Error occurred], unable to send suspicious activity event with ID: ${eventIdentifier} to Zendesk, ${
-        (error as Error).message
-      }`
-    );
     throw new Error(
-      `[Error occurred], unable to send suspicious activity event with ID: ${eventIdentifier} to Zendesk, ${
+      `Unable to send suspicious activity event with ID: ${eventIdentifier} to Zendesk, ${
         (error as Error).message
       }`
     );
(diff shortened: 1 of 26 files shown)


src/create-support-ticket.ts
```diff
@@ -189,13 +189,8 @@ export const handler = async (
     console.log(`finished processing event with ID: ${eventIdentifier}`);
     return input;
   } catch (error) {
-    console.error(
-      `[Error occurred], unable to send suspicious activity event with ID: ${eventIdentifier} to Zendesk, ${
-        (error as Error).message
```
peterfajemisincabinetoffice, line 194: Found this console.error useful as part of a recent debugging activity. Why is it removed pls?

peterfajemisincabinetoffice: Can we have a quick run through of the changes in this PR pls when you are free

peterfajemisincabinetoffice (approved):

PR #375, opened 2024-09-18
OLH-1321 - Performance testing: Throttling in several Lambda function

peterfajemisincabinetoffice:
OLH-1321 - Performance testing: Throttling in several Lambda function during load tests

## Proposed changes
Add ProvisionedConcurrency to the identified Lambda so that a number of pre-initialized execution environments are allocated to the function. Reduces cold start latency and should stop the throttling.

### What changed

Update cloudformation template to configure provisioned concurrency

### Why did it change

So that cold start are reduced and consequently stops throttling in lamnbdas when under load.

## Testing

Deploy to dev environment
Run performance test
Check Cloudwatch metrics and confirm no throttling is happening under load.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -755,6 +755,7 @@ Resources:
         TargetArn: !GetAtt SaveRawEventsDeadLetterQueue.Arn
       Handler: save-raw-events.handler
       Role: !GetAtt SaveRawEventsRole.Arn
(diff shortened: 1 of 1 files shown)


latifcabinetoffice:
Looks good. Is it worth getting perf test team to generate a report which will indicate what cost difference we can expect with this configuration?

latifcabinetoffice (approved):

PR #439, opened 2024-11-18
OLH-2228 - Write a pre-merge test for a real use case of OLH

peterfajemisincabinetoffice:
## Proposed changes

OLH-2228 - Write a pre-merge test for a real use case of OLH

### What changed

Update Python script to include test for the writing an activity log and checking the activity log is recorded in dynamo.

### Why did it change

So that we get pre-merge testing to do real regression test against the one login home service

- [x] Added new secret or systems manager parameter
- [x] Added new environment variable

### Permissions

## Testing

Deploy to demo environment, trigger pre-merge testing and confirm it runs successfully

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -1,6 +1,9 @@
-name: pre-merge check
+name: Pre Merge Check
 
 on:
+  push:
+    branches:
+      - OLH-2228
   workflow_call:
     inputs:
       gitRef:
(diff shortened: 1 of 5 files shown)


.github/workflows/pre-merge.yaml
```diff
@@ -1,43 +1,116 @@
-name: pre-merge check
+name: Pre Merge Check
 
 on:
+  push:
+    branches:
+      - OLH-2228
```
saralk, line 6: will this need to be removed before merge?

peterfajemisincabinetoffice: Thanks @saralk , leaving as is for now as it doesn't impact anything and part of the required code when we workflow block the pipeline, we would just change the branch to the main branch.

saralk (approved):

PR #608, opened 2025-04-15
Olh 2404 warn if unexpected txma client

latifcabinetoffice:
## Proposed changes

[OLH-2404] Trigger an Alarm if we receive a unexpected TXMA Event Client ID

### What changed

Added a check against the rp registry for the txma client ID.
If the ID doesn't match any of the IDs in the client registry, it logs a warning.
This warning is picked up by a log filter and in turn raises an alarm. 
This will let us know if there is a unexpected txma client ID event and potentially allow us to quickly respond to bad data entering our databases.

### Why did it change

Improve monitoring for unexpected client ids being processed to our databases.

## Testing

Deployed and tested in Dev.

[OLH-2404]: https://govukverify.atlassian.net/browse/OLH-2404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8184,7 +8184,7 @@
     "node_modules/di-account-management-rp-registry": {
       "name": "di-account-management-client-registry",
       "version": "1.0.0",
-      "resolved": "git+ssh://git@github.com/govuk-one-login/di-account-management-rp-registry.git#19beeb5dd992640eb084812ff951f4745affb8c7",
+      "resolved": "git+ssh://git@github.com/govuk-one-login/di-account-management-rp-registry.git#e95395952f62ba21cf04764e037aba1634d5ffd4",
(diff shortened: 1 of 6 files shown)


alex9smith (changes requested):
Could you add tests for this log message please? I think we need one that asserts there's a warning when the ID's not known and one that there's no warning when we do know the client ID. 

Also we should migrate all these lambdas to use the logger from the AWS lambda powertools package since it includes lots of helpful things like injecting the lambda context and trace IDs. I'll write up a separate ticket for that though - no need to tackle it in this PR.

alex9smith (approved):
Thanks!

PR #981, opened 2026-05-07
OLH-4066: migrate to ESM

liegeandlief:
This PR includes all the changes necessary to migrate from CommonJS to ESM.

The major part of the migration was migrating tests from Jest to Vitest. This was reasonably straightforward due to their similar APIs. There were a couple of places where the source files had to be changed in order for the tests to work with Vitest. The changes required that code which is mocked in the tests be moved into separate files. There is no change to the actual logic. The files in question are:

- `src/notification-service.ts` (code moved into `src/notification-service-utils.ts` and `src/notification-service-client.ts`)
- `src/delete-email-subscriptions.ts` (code moved into `src/delete-email-subscriptions-utils.ts`)

The other main change required for the migration was to ensure all imports end with `.js`.


src/tests/delete-email-subscriptions.test.ts
```diff
@@ -1,45 +1,60 @@
...
 
 describe("handler", () => {
   beforeEach(() => {
-    jest.restoreAllMocks();
+    vi.clearAllMocks();
+    validateUserDataMock.mockImplementation((data) => data);
+    deleteEmailSubscriptionMock.mockResolvedValue(undefined);
     process.env.AWS_REGION = "AWS_REGION";
+    process.env.GOV_ACCOUNTS_PUBLISHING_API_TOKEN = "test-token";
+    process.env.GOV_ACCOUNTS_PUBLISHING_API_URL = "https://api.example.com";
```
saralk, line 53: should we restore the original value of these variables in the afterEach block

saralk (approved):

PR #1034, opened 2026-06-08
OLH-3559: Activity log table analysis Lambda

mattvot:
### What changed

Adds a manually-invocable Lambda that scans the `activity_log` DynamoDB table using parallel segments and produces a JSON report with:
- Per-user item count distribution (mean, median, percentiles)
- Concentration metrics (top 1%/5%/10% of users' share of items)
- User bucketing by item count (1, 2-5, 6-10, ... 100000+)
- Items by age range (0-1 months through 24+ months)
- TTL impact simulation for 3/6/12/18/24 month thresholds

### Why did it change

The `activity_log` table has no TTL and grows unboundedly. We need data to assess retention strategies (time-based TTL vs segregation by value). This Lambda produces the analysis needed to make that decision.

### Example output

```json
{
  "scan_date": "2026-06-08T12:00:00.000Z",
  "total_items": 1843927,
  "total_users": 241000,
  "scan_duration_seconds": 47,
  "items_per_user_distribution": {
    "mean": 7.6,
    "median": 3,
    "p75": 8,
    "p90": 14,
    "p95": 22,
    "p99": 61,
    "max": 3870
  },
  "concentration": {
    "top_1_pct_users_own_pct_of_items": 31,
    "top_5_pct_users_own_pct_of_items": 55,
    "top_10_pct_users_own_pct_of_items": 71
  },
  "items_per_user_buckets": {
    "1": { "user_count": 58200, "total_items": 58200 },
    "2-5": { "user_count": 97400, "total_items": 312000 },
    "6-10": { "user_count": 43100, "total_items": 321000 },
    "11-25": { "user_count": 28600, "total_items": 412000 },
    "26-50": { "user_count": 9200, "total_items": 298000 },
    "51-100": { "user_count": 3100, "total_items": 214000 },
    "101-500": { "user_count": 1180, "total_items": 178000 },
    "501-1000": { "user_count": 142, "total_items": 89000 },
    "1001-10000": { "user_count": 38, "total_items": 62000 },
    "10001-100000": { "user_count": 0, "total_items": 0 },
    "100000+": { "user_count": 0, "total_items": 0 }
  },
  "items_by_age_bucket": {
    "0-1_months": { "count": 231000 },
    "1-3_months": { "count": 389000 },
    "3-6_months": { "count": 341000 },
    "6-12_months": { "count": 462000 },
    "12-18_months": { "count": 243000 },
    "18-24_months": { "count": 112000 },
    "24+_months": { "count": 65900 }
  },
[...]

PR #1037, opened 2026-06-10
OLH-3559: Self-invoking scan for large table analysis

mattvot:
### What changed

The analysis Lambda now self-invokes to handle tables that exceed the 15-minute Lambda timeout. Instead of scanning the entire table in one execution, it:

1. Scans for 13 minutes across all segments in parallel
2. Stops at the deadline, preserving scan position and in-progress user state via cursors
3. Converts per-user counters into compact frequency maps (ChunkResult) that accumulate across invocations
4. Async-invokes itself with the accumulated state and cursors for incomplete segments
5. On the final invocation, computes the report from the merged frequency maps

The frequency-map approach keeps the payload bounded (~few KB) regardless of user count while producing exact percentiles, concentration metrics, and TTL simulations.

### Why did it change

The Lambda timed out at 15 minutes having scanned about a third of the items needed in production. This change enables it to complete scans of any table size within the Lambda execution model.

### Related links

- Previous PR: #1034 (original Lambda)

### Permissions

Adds `lambda:InvokeFunction` scoped to the function's own ARN (self-invocation only).

### Monitoring

### DIFF
--- a/src/analyse-activity-log/chunk-result.ts
+++ b/src/analyse-activity-log/chunk-result.ts
@@ -1,0 +1,108 @@
+import { CounterIndex } from "./age-thresholds.js";
+
+export interface ChunkResult {
+  totalCountFrequency: Record<number, number>;
+  ageBuckets: number[];
+  ttlRetainedFrequency: Record<string, Record<number, number>>;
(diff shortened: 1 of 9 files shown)


martin-gwar (approved):
Looks good to me!

PR #1058, opened 2026-06-17
OLH-4305 add inactive account tracker deletion lambda

saralk:
### What changed

Create a lambda to delete data for a specified user from the inactive account tracker table (enabled only in development)

subscribe that lambda to to UserAccountDeletion SNS topic

### Why did it change

This is required as we will store user data which needs to be removed when the user delete's their account 

### Permissions

### Monitoring

## Testing

Deployed the branch to dev and manually triggered an event on the SNS topic

### DIFF
--- a/src/delete-inactive-account-tracker.ts
+++ b/src/delete-inactive-account-tracker.ts
@@ -1,0 +1,94 @@
+import { Context, SNSEvent } from "aws-lambda";
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
+import {
+  DynamoDBDocumentClient,
+  QueryCommand,
+  DeleteCommand,
+} from "@aws-sdk/lib-dynamodb";
+import { UserData } from "./common/model.js";
+import { getEnvironmentVariable } from "./common/utils.js";
+import { Logger } from "@aws-lambda-powertools/logger";
+
+const logger = new Logger();
+
+const marshallOptions = {
+  convertClassInstanceToMap: true,
+};
+const translateConfig = { marshallOptions };
+
+const dynamoClient = new DynamoDBClient({});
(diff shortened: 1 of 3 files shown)


src/delete-inactive-account-tracker.ts
```diff
@@ -0,0 +1,94 @@
...
+    )
+  );
+};
+
+export const handler = async (
+  event: SNSEvent,
+  context: Context
+): Promise<void> => {
+  logger.addContext(context);
+  await Promise.all(
```
liegeandlief: Would `allSettled` work better here?

src/delete-inactive-account-tracker.ts
```diff
@@ -0,0 +1,94 @@
...
+      ExpressionAttributeValues: { ":uid": userData.user_id },
+    })
+  );
+
+  if (!queryResponse.Items || queryResponse.Items.length === 0) {
+    logger.info("no inactive account tracker records found for user");
+    return;
+  }
+
+  await Promise.all(
```
liegeandlief: Same here re. allSettled

saralk: for what reason?

liegeandlief: as it currently stands if any of the promises rejects then the promise.all will abort and reject even if other promises are still pending? Not sure if that is the intended behaviour

saralk: in that scenario, we'd eat the error I think? By using `all` the lambda would fail, and the item would end up in the DLQ

liegeandlief: Makes sense!

liegeandlief (approved):

PR #1063, opened 2026-06-17
OLH-4239 Create account processer lambda

saralk:
Introduces a new scheduled Lambda — QueryAndDispatchInactiveAccounts — that queries a DynamoDB table for inactive accounts due for deletion and dispatches them to SQS queues for downstream processing (notifications/deletion).

### DIFF
--- a/src/common/model.ts
+++ b/src/common/model.ts
@@ -210,7 +210,7 @@ export interface Personalisation {
   timeCy: string;
 }
 
-export type InactiveAccountStatus = "pending" | "deleting" | "30DayWarningSent" | "7DayWarningSent" | "permenantSuspension"
+export type InactiveAccountStatus = "pending" | "deleting" | "30DayWarningSent" | "7DayWarningSent" | "permanentSuspension"
 
 export interface InactiveAccountTrackerRecord {
   dateForDeletion: string;
--- a/src/common/model.ts
+++ b/src/common/model.ts
@@ -210,7 +210,7 @@ export interface Personalisation {
   timeCy: string;
 }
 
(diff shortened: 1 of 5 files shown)


src/common/model.ts
```diff
@@ -210,7 +210,7 @@ export interface Personalisation {
   timeCy: string;
 }
 
-type InactiveAccountStatus = "pending" | "deleting"
+export type InactiveAccountStatus = "pending" | "deleting" | "30DayWarningSent" | "7DayWarningSent" | "permenantSuspension"
```
liegeandlief: Nit: spelling mistake `permenant` -> `permanent`

template.yaml
```diff
@@ -3174,6 +3168,27 @@ Resources:
...
+    DeletionPolicy: Delete
+    UpdateReplacePolicy: Delete
+
+  AccountDeletionQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      KmsMasterKeyId: !GetAtt QueueKmsKey.Arn
+    DeletionPolicy: Delete
+    UpdateReplacePolicy: Delete
+
```
liegeandlief, line 3191: Do these queues need encryption at rest?

liegeandlief (approved):
A couple of small comments but otherwise LGTM

template.yaml
saralk: By added a KmsMasterKeyId, that encrypts at rest, I've verified in the console that it is encrypted

liegeandlief: Oh yeah I completely missed that!

PR #1084, opened 2026-07-07
OLH-4412: Include user email in inactive account tracker record

danacotoran:
## Proposed changes

Update the inactive account tracker lambda to include the user's email with inactive account tracker record.

### DIFF
--- a/src/update-inactive-account-tracker.ts
+++ b/src/update-inactive-account-tracker.ts
@@ -59,13 +59,15 @@ export const handler = async (
     const txmaEvent = unmarshall(
       record.dynamodb?.NewImage?.event.M as Record<string, AttributeValue>
     ) as TxmaEvent;
-
+    
     const userId = txmaEvent.user?.user_id;
-    const email = txmaEvent.user?.email;
     assert(userId !== undefined, "user_id is undefined in the event");
-    assert(email !== undefined, "email is undefined in the event");
-
+    if (txmaEvent.user?.email === undefined) {
+      logger.warn(`AUTH_EVENT_NO_EMAIL for userId ${userId}`)
+    }
+    
     const currentTrackerRecord = await getCurrentRecordForUser(userId, tableName);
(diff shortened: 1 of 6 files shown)


src/update-inactive-account-tracker.ts
```diff
@@ -61,7 +61,9 @@ export const handler = async (
     ) as TxmaEvent;
 
     const userId = txmaEvent.user?.user_id;
+    const email = txmaEvent.user?.email;
     assert(userId !== undefined, "user_id is undefined in the event");
+    assert(email !== undefined, "email is undefined in the event");
```
saralk: I think the ticket wasn't clear (my fault!), but we should still record the activity, and leave the email field empty (or use the one that already exists in the table). But still trigger an alarm

danacotoran: Oooh got it, thanks! I also shouldn't have assumed to be fair.

So I understand it correctly,:
- if there is no email on the data on the event, but an email was recorded for that user previously, we should use that
- if there is no email on the data on the event, and an email is also not specified in the table, we should leave it blank
- for both scenarios above, an alarm should be triggered (clarification: what do you mean by this, more specifically?)

saralk: Yep, that's correct.

For the alarm, I mean an alarm in CloudWatch, so we'd get observability on it (since this shouldn't happen).

You can either create a metric and then define an alarm that triggers when there is a datapoint.

Or (imo the simpler method), you can write a warning to the log and then add an alarm that triggers based on that (this is an example of the [logging](https://github.com/govuk-one-login/di-account-management-backend/blob/aded1e33c53384c1969dfa35c8d7450386ae5796/src/update-inactive-account-tracker.ts#L69-L71) and [alarm](https://github.com/govuk-one-login/di-account-management-backend/blob/aded1e33c53384c1969dfa35c8d7450386ae5796/template.yaml#L3340-L3349))

saralk (approved):

PR #1101, opened 2026-07-17
OLH-4471: Retry on delete email subscriptions failure

danacotoran:
### What changed
Add retries to delete email subscription call before allowing lambda to error.

### Why did it change
Logging has revealed that this method has previously failed a few times, possibly due to GOVUK Api flakiness. This should add a bit more resilience to the email subscription delete lambda.

### DIFF
--- a/src/delete-email-subscriptions.ts
+++ b/src/delete-email-subscriptions.ts
@@ -8,6 +8,24 @@ import {
 
 const logger = new Logger();
 
+const retryDeleteEmailSubscription = async (
+  userData: UserData
+): Promise<void> => {
+  const retries = 3;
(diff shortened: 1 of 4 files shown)


src/delete-email-subscriptions.ts
```diff
@@ -8,6 +8,24 @@ import {
 
 const logger = new Logger();
 
+const retryDeleteEmailSubscription = async (
```
liegeandlief: Could this be made into a generic retry function in case we want to use it elsewhere in future?

danacotoran: Good shout, will do!

liegeandlief (approved):
Very nice!

PR #1104, opened 2026-07-22
BAU enable IAD in prod

saralk:
Enable IAD in all environments

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -3159,7 +3159,6 @@ Resources:
           Properties:
             BatchSize: 1
             Stream: !GetAtt RawEventsStore.StreamArn
-            Enabled: !If [IsDevelopment, true, false]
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -4126,7 +4125,6 @@ Resources:
   ####################################
 
   DeleteUserNotificationsSubscription:
-    Condition: IsDevelopment
```
mattvot, line 4129: Doesn't `DeleteUserNotifications*` related resources do the actual deletion, not tracking.

mattvot (changes requested):

template.yaml
saralk: this is so that when a user deletes their account, their data is removed from the tracking table

mattvot (approved):

PR #1106, opened 2026-07-23
OLH-4455: Publish daily deletion forecast metrics

martin-gwar:
## Proposed changes

[OLH-4455]: Publish daily deletion forecast metrics

### What changed?

Created a scheduled Lambda thats fired at 6am each day that queries the InactiveAccountTrackerStore table, counts items in each of the next 180 dateForDeletion partitions, and publishes those counts as a CloudWatch custom metric with a dimension per date.
- Generates array of 180 future dates ["2026-07-22", "2026-07-23", ... "2026-01-17"]
- queries InactiveAccountTrackerStore table for counts for each of the 180 future dates in parallel
- publish metrics Namespace: "account-management-backend"; MetricName: "InactiveAccountsScheduledForDeletion"

### Permissions

### Monitoring

## How to review

[OLH-4455]: https://govukverify.atlassian.net/browse/OLH-4455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -505,7 +505,6 @@
       "resolved": "https://registry.npmjs.org/@aws-sdk/client-secrets-manager/-/client-secrets-manager-3.1081.0.tgz",
       "integrity": "sha512-rPLag4HuZYW/vCUMGZ19zFWgi8Uw/rARgxSLmPpX2ua9h1ysXX+KBwf4/o74uJO5qanVPHJv59aZ03/WBVrQKw==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@aws-sdk/core": "^3.974.29",
         "@aws-sdk/credential-provider-node": "^3.972.64",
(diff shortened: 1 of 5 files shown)


src/inactive-account-deletion-forecast.ts
```diff
@@ -0,0 +1,84 @@
...
+  );
+
+  const metrics: MetricDatum[] = dates.map((date, i) => ({
+    MetricName: METRIC_NAME,
+    Dimensions: [{ Name: "DateForDeletion", Value: date }],
+    Value: counts[i],
+    Unit: "Count",
+  }));
+
+  await publishMetrics(metrics);
```
liegeandlief: We already have a pattern for publishing metrics in this repo which doesn't need `@aws-sdk/client-cloudwatch`. I think it makes use of lambda powertools and you just need to call `initMetrics` in the module scope. If possible we should stick with the existing pattern rather than having multiple different ways of doing things

package-lock.json
liegeandlief, line 1: Is there a reason this file has changed? If not then can it be removed?

template.yaml
```diff
@@ -7120,6 +7127,119 @@ Resources:
...
         - !Sub "{{resolve:ssm:/${AlertingStackName}/SNS/LowAlertNotificationTopic/ARN}}"
 
+  #########################################
+  # Inactive Account Deletion Forecast
+  #########################################
+
+  InactiveAccountDeletionForecastFunction:
+    Type: AWS::Serverless::Function
+    DependsOn:
+      - InactiveAccountDeletionForecastFunctionLogGroup
```
liegeandlief, line 7137: Not sure this is necessary, CloudFormation should be able to work this out itself

liegeandlief (approved):
A couple of tiny things but otherwise looks good to me 👍

PR #1116, opened 2026-07-30
OLH-4529: save raw events partial batch processing

liegeandlief:
Updates the save raw events function to support partial batch processing. This means that only failed messages are retried, not the whole batch.

Increased the batch size to 100 and set the function timeout to 30 seconds.

This has been tested by deploying to dev and manually testing the lambda in the AWS console. I can confirm I have seen successful responses where `batchItemFailures` is an empty array because all messages were processed. I have also seen unsuccessful responses where `batchItemFailures` contains items because there were bad events in the payload.

### DIFF
--- a/src/save-raw-events.ts
+++ b/src/save-raw-events.ts
@@ -1,4 +1,4 @@
-import { Context, SQSEvent } from "aws-lambda";
+import { Context, SQSBatchResponse, SQSEvent } from "aws-lambda";
 import crypto from "node:crypto";
 import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
 import {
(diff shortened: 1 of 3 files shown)


mattvot:
To do the batch size of 100 in 30 seconds gives 300ms per event to process and store it.

If we time out on a large batch, the whole lambda fails and the whole batch goes back for another attempt.

I've not seen myself yet how long it takes. 300ms per event is hopefully plenty. Do we have evidence that this is sufficient time in prod?

mattvot:
Looked at the prod logs and seen we're looking in the 10s of milliseconds range, not 100s. Happy to approve and merge

mattvot (approved):

liegeandlief:
> Looked at the prod logs and seen we're looking in the 10s of milliseconds range, not 100s. Happy to approve and merge

The records are processed in parallel not sequentially so 30 seconds for 100 records should be plenty, although it may be worth checking and tweaking the memory of the lambda

PR #1118, opened 2026-07-30
OLH-4542: Add more detail to save-raw-events errors

danacotoran:
### What changed

Add detail about the field or fields which are missing or invalid to the validate user and validate txma event body functions, in order to facilitate future debugging if/when validation fails.

## How to review
Suggestions to further improvements are welcomed.

### DIFF
--- a/src/save-raw-events.ts
+++ b/src/save-raw-events.ts
@@ -45,7 +45,15 @@ export const validateUser = (user: UserData): void => {
       typeofUserUserId: typeof user.user_id,
       typeofUserSessionId: typeof user.session_id,
     });
-    throw new Error("Could not validate User");
+    const missingFields: string[] = [];
+
+    if (!user.user_id) {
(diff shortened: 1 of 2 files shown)


src/save-raw-events.ts
```diff
@@ -45,7 +45,15 @@ export const validateUser = (user: UserData): void => {
...
-    throw new Error("Could not validate User");
+    const missingFields: string[] = [];
+
+    if (!user.user_id) {
+      missingFields.push(`user_id is ${user.user_id}`);
+    }
+    if (!user.session_id) {
+      missingFields.push(`session_id is ${user.session_id}`);
+    }
+    throw new Error(`Could not validate User: ${missingFields.join(", ")}`);
```
mattvot: When this throws, we won't have the event_name or event_id. Can we include those in this log please?

mattvot (changes requested):

mattvot (approved):

PR #1122, opened 2026-07-31
OLH-4527: Historic millisecond timestamp workaround

danacotoran:
### What changed

Some txma events may have had timestamps recorded in milliseconds when they should have been in seconds. Detect if the date is in milliseconds by checking the number of digits it is made up of. Starting from 2001, Unix timestamps in milliseconds will have been 13 digits long.


liegeandlief (approved):
Nice simple solution 👍

PR #1141, opened 2026-08-06
OLH-4536: Only update email address if it's the latest

danacotoran:
### What changed
Introduce logic to prevent accidentally reverting a user's tracked email address to an older version, in the situation where events are sent/processed out of order.

When an event comes through with a timestamp that is earlier than the timestamp of when an email was last saved, the newest email address should be kept.

### DIFF
--- a/src/tests/update-inactive-account-tracker.test.ts
+++ b/src/tests/update-inactive-account-tracker.test.ts
@@ -380,7 +380,7 @@ describe("UpdateInactiveAccountTracker handler", () => {
 
   test("updates emailAddressSource, emailAddressSourceId and emailAddressLastUpdated when email changes", async () => {
     dynamoMock.on(QueryCommand).resolves({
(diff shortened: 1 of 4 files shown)


liegeandlief (commented):
I'm really struggling to understand all the logic in this file now. It seems like there is duplication of logic and all sorts of confusing stuff going on. I think it needs to be refactored to be more understandable and so we can be sure it is doing what we expect it to.

danacotoran:
> struggling to understand all the logic in this file now 

Thanks @liegeandlief, happy to add some comments/refactor but just to clarify, which file are you referring to? 
I am guessing you're talking about `update-inactive-account-tracker.ts` since the changes in the other one are fairly straightforward (to me, anyway 😅) but I don't really want to assume.

liegeandlief:
> > struggling to understand all the logic in this file now
> 
> Thanks @liegeandlief, happy to add some comments/refactor but just to clarify, which file are you referring to? I am guessing you're talking about `update-inactive-account-tracker.ts` since the changes in the other one are fairly straightforward (to me, anyway 😅) but I don't really want to assume.

@danacotoran yes it's `update-inactive-account-tracker.ts` I'm having trouble following, I think it could be simplified a fair bit

